PR #9, opened 2018-10-18
Removing use of `BaseObject` constructor

larryli:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no

### DIFF
--- a/src/DataReader.php
+++ b/src/DataReader.php
@@ -68,7 +68,6 @@ class DataReader extends \yii\base\BaseObject implements \Iterator, \Countable
     {
         $this->_statement = $command->pdoStatement;
         $this->_statement->setFetchMode(\PDO::FETCH_ASSOC);
-        parent::__construct($config);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/DataReader.php
```diff
@@ -68,7 +68,6 @@ public function __construct(Command $command, $config = [])
     {
         $this->_statement = $command->pdoStatement;
         $this->_statement->setFetchMode(\PDO::FETCH_ASSOC);
-        parent::__construct($config);
```
cebe: you still have `$config` in the constructor signature.

Btw, can someone point me to the discussion, on why this has been removed from Yii core?

larryli: I don't know.
Just fix "Error: Cannot call constructor".

kids-return:
Hi @cebe 
The reason is here
https://github.com/yiisoft/yii-core/commit/a16ba215994336fcc50e7340824af3e102df7fe2#diff-828ad3929c2c6e3ba47c513ec3625631

machour:
Closing this as it was handled in https://github.com/yiisoft/db/commit/41fafcfd606deb453988112c5ec2786793487b2d

PR #26, opened 2018-11-13
Added dns example

kids-return:
Added dns example
http://php.net/manual/en/pdo.construct.php
http://php.net/manual/en/ref.pdo-mysql.connection.php
https://github.com/yiisoft/db/pull/13


kids-return:
see:https://github.com/yiisoft/db/pull/13

samdark:
Merged. Thanks!

PR #32, opened 2018-12-27
Fix namespaces

machour:
Fixed Exceptions namespaces and some other minor stuff.
PR needed to be able to work on db-* tests without hassle.

### DIFF
--- a/tests/unit/ConnectionTest.php
+++ b/tests/unit/ConnectionTest.php
@@ -8,9 +8,9 @@
 namespace yii\db\tests\unit;
 
 use yii\helpers\Yii;
-use yii\base\InvalidConfigException;
-use yii\caching\ArrayCache;
-use yii\caching\Cache;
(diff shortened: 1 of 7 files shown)


tests/unit/ConnectionTest.php
```diff
@@ -461,14 +461,14 @@ public function testDSNConfig()
...
+        $this->expectException('yii\exceptions\InvalidConfigException');
         $connection = new Connection(['dsn' => $dsn]);
     }
 
     public function testServerStatusCacheWorks()
     {
-        $cache = new Cache(['handler' => new ArrayCache()]);
-        Yii::getApp()->set('cache', $cache);
+        $cache = new Cache(new ArrayCache());
+        $this->container->set('cache', $cache);
```
machour: @hiqsol is this correct ? Previously this was setting `cache` in the Application, but now it's doing it on `$this`
(yes, I'm still confused 😅 )

PR #34, opened 2018-12-28
Fix missing query constructor

FabrizioCaldarelli:
Fix missing yii\db\Query object that cause no table use error in tests.

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -1301,6 +1301,27 @@ PATTERN;
     }
 
     /**
+     * Constructor for query object
+     */
+    public function __construct($config = [])
+    {
+        $this->where = isset($config['where'])?$config['where']:null;
+        $this->limit = isset($config['limit'])?$config['limit']:null;
+        $this->offset = isset($config['offset'])?$config['offset']:null;
+        $this->orderBy = isset($config['orderBy'])?$config['orderBy']:null;
+        $this->indexBy = isset($config['indexBy'])?$config['indexBy']:null;
+        $this->select = isset($config['select'])?$config['select']:null;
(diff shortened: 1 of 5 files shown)


src/Query.php
```diff
@@ -1300,6 +1300,27 @@ public static function create($from)
         ]);
     }
 
+    /**
+     * Constructor for query object
+     */
+    public function __construct($config = [])
+    {
```
hiqsol: please use `yii\di\AbstractContainer::configure()` instead

FabrizioCaldarelli: I've searching for some example, but I've not found nothing that use AbstractContainer.

I've tryied with:

        return \yii\di\AbstractContainer::configure( Query::class, $config);

but the first parameters has to be an object. 

I can't pass new Query() because I'm already inside the Query contructor

hiqsol: Just use `$this` it is available in the constructor.
You can see example here: https://github.com/yiisoft/yii-core/blob/d87cecdb27b21773e75c08137fe644114823f8d8/src/validators/Validator.php#L197

hiqsol:
Thank you!

src/Query.php
```diff
@@ -1300,6 +1300,16 @@ public static function create($from)
         ]);
     }
 
+    /**
+     * Constructor for query object
+     */
+    public function __construct($config = [])
```
rob006:
```suggestion
    public function __construct(array $config = [])
```

hiqsol: Good point. Done.

hiqsol: Thank you!

Issue #40, opened 2018-12-31
Command testBindParamValue fails


FabrizioCaldarelli:
Here $float_val = 1.23, instead db column type is defined as double(4,3), so its value is 1.230.

https://github.com/yiisoft/db/blob/1351c16e684a6f1858b9034fdac6ad9ef3fc231c/tests/unit/CommandTest.php#L207

When checked equality, test fails.

FabrizioCaldarelli:
According to:

If I change that line to:
https://phpunit.de/manual/6.5/en/appendixes.assertions.html

`        
$this->assertEquals((float)$floatCol, (float) $row['float_col']);
`

test succeded.

Are you agree to update that line?

machour:
If that test was copied over from yii2 and was working without casting to float, we better understand what changed and why. 

Could you check how it's done in yii2?

FabrizioCaldarelli:
Explained here: https://github.com/yiisoft/db/pull/44#pullrequestreview-188864375

PR #44, opened 2019-01-02
Fix tests

FabrizioCaldarelli:


tests/unit/CommandTest.php
```diff
@@ -204,7 +204,7 @@ public function testBindParamValue()
         $row = $command->queryOne();
         $this->assertEquals($intCol, $row['int_col']);
         $this->assertEquals($charCol, $row['char_col']);
-        $this->assertEquals($floatCol, $row['float_col']);
+        $this->assertEquals((float)$floatCol, (float)$row['float_col']);
```
machour: Copied from #40: 

> If that test was copied over from yii2 and was working without casting to float, we better understand what changed and why.
>
> Could you check how it's done in yii2?

machour: The same test is present here for yii2:
   https://github.com/yiisoft/yii2/blob/master/tests/framework/db/CommandTest.php#L210

MySQL 5.7 is used for yii2: https://github.com/yiisoft/yii2/blob/master/.travis.yml#L55
(this can also be seen in the logs: )

machour: (You may want to compare data sets used for tests between yii2 & yii3)

FabrizioCaldarelli: I found the problem.

Sebatian comparator package has changed this line:

https://github.com/sebastianbergmann/comparator/blob/33825cd25f073faafe98c4a017a5d2b69f3e09d3/src/ScalarComparator.php#L64

at this commit:

https://github.com/sebastianbergmann/comparator/commit/c4e6d5c693d4dce8f3db478f09ef058b1b9c987e

So, Yii2 does a "!=" check, instead Yii3 does a "!==" check, that fails, because db field is "1.230", instead floatval is "1.23".

So I think that my PR is valid.

rob006: @FabrizioCaldarelli  This should not be the case, since `$floatCol` should be either float or int in this test.

FabrizioCaldarelli: @rob006 Exactly, $floatCol is a float (1.23), but $row['float_col'] is a string, returned by $query->one(), with "1.230" value.

Since Sebastian comparator packages changed check from "!=" to "!==", now the two values don't match.

tests/unit/CommandTest.php
```diff
@@ -1455,12 +1455,14 @@ public function testCreateView()
...
-        if ($db->getSchema()->getTableSchema('testCreateViewTable')) {
-            $db->createCommand()->dropTable('testCreateViewTable')->execute();
-        }
+        // Order of the next two "if" conditions is very important, because testCreateView depends on testCreateViewTable.
+        // If testCreateViewTable was deleted before testCreateView, at the second round this test will always fail becase testCreateView exists but it is referred to not existing table
         if ($db->getSchema()->getTableSchema('testCreateView') !== null) {
             $db->createCommand()->dropView('testCreateView')->execute();
         }
+        if ($db->getSchema()->getTableSchema('testCreateViewTable')) {
+            $db->createCommand()->dropTable('testCreateViewTable')->execute();
```
machour: Just like for float casting, I'd like to understand why is this change needed if it used to work on yii2.

tests/unit/validators/UniqueValidatorTest.php
```diff
@@ -231,7 +231,7 @@ public function testValidateScopeNamespaceTargetClassForNewClass()
         $validator = new UniqueValidator();
 
         /** @var Profile $profileModel */
-        $profileModel = new Profile(['description' => 'profile customer 1']);
+        $profileModel = Yii::createObject(['__class' => Profile::class, 'description' => 'profile customer 1']);
```
machour: Why not use $this->factory, like above?

And globally, why use the factory when a simple instantiation could be enough. Is there an init() being called? (On phone right now, can't check)

tests/unit/CommandTest.php
rob006: @FabrizioCaldarelli But [this condition](https://github.com/sebastianbergmann/comparator/blob/33825cd25f073faafe98c4a017a5d2b69f3e09d3/src/ScalarComparator.php#L64) will pass only if both arguments are strings, so not in this case. And then [this condition](https://github.com/sebastianbergmann/comparator/blob/33825cd25f073faafe98c4a017a5d2b69f3e09d3/src/ScalarComparator.php#L75) will be checked, but it should return `false`: https://3v4l.org/eui82 . 

So definitely something weird is happening here.

tests/unit/CommandTest.php
machour: I was saying: If this is a test copied from yii2, and it used to work with yii2, but not anymore with yii3, the tests should not be changed. 

Instead, we should look for the reason it's failing now, cause this may be caused by a new bug.

Can you check what's different between yii2 and yii3 here?

tests/unit/CommandTest.php
FabrizioCaldarelli: @rob006 This is caused by PDOStatement::bindParam. At beginning $floatVal is a float, but after executing bindParam: 
https://github.com/yiisoft/db/blob/54e7a7ef9f7683bb641a13574c5c34cbd8c5dc61/src/Command.php#L298

PDOStatement::bindParam will convert numeric value in string, as described here: https://github.com/yiisoft/db/blob/54e7a7ef9f7683bb641a13574c5c34cbd8c5dc61/src/Command.php#L298

tests/unit/validators/UniqueValidatorTest.php
FabrizioCaldarelli: I preferred to keep previous syntax, using array to pass properties values.

Do you want that I change it to explicit version?

tests/unit/CommandTest.php
FabrizioCaldarelli: Surely, when Travis starts the test, its database not exists, so this code change is unuseful.

It is useful if you start from existing database (if you have already executed the test), because the testCreateViewTable is not dropped by mysql.sql.

Try.

FabrizioCaldarelli: This is what happens using previous code version:

```bash
root@ecf9af83c6f8:/app/web/db-mysql# echo "create database yiitest" | mysql -h mysql -u travis -ptravis
root@ecf9af83c6f8:/app/web/db-mysql# vendor/bin/phpunit tests/unit/CommandTest.php --filter testCreateView
PHPUnit 7.5.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.3 with Xdebug 2.6.0-dev
Configuration: /app/web/db-mysql/phpunit.xml.dist

.                                                                   1 / 1 (100%)

Time: 1.4 seconds, Memory: 4.00MB

OK (1 test, 1 assertion)
root@ecf9af83c6f8:/app/web/db-mysql# vendor/bin/phpunit tests/unit/CommandTest.php --filter testCreateView
PHPUnit 7.5.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.3 with Xdebug 2.6.0-dev
Configuration: /app/web/db-mysql/phpunit.xml.dist

E                                                                   1 / 1 (100%)

Time: 1.56 seconds, Memory: 4.00MB

There was 1 error:

1) yii\db\mysql\tests\CommandTest::testCreateView
yii\db\Exception: SQLSTATE[HY000]: General error: 1356 View 'yiitest.testCreateView' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them
The SQL being executed was: SHOW FULL COLUMNS FROM `testCreateView`

/app/web/db-mysql/vendor/yiisoft/db/src/Schema.php:669
/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:1263
/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:1148
/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:399
/app/web/db-mysql/src/Schema.php:313
/app/web/db-mysql/src/Schema.php:126
/app/web/db-mysql/vendor/yiisoft/db/src/Schema.php:749
/app/web/db-mysql/vendor/yiisoft/db/src/Schema.php:199
/app/web/db-mysql/vendor/yiisoft/db/tests/unit/CommandTest.php:1463
/app/web/db-mysql/vendor/phpunit/phpunit/phpunit:61

Caused by
PDOException: SQLSTATE[HY000]: General error: 1356 View 'yiitest.testCreateView' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them

/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:1258
/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:1148
/app/web/db-mysql/vendor/yiisoft/db/src/Command.php:399
/app/web/db-mysql/src/Schema.php:313
/app/web/db-mysql/src/Schema.php:126
/app/web/db-mysql/vendor/yiisoft/db/src/Schema.php:749
/app/web/db-mysql/vendor/yiisoft/db/src/Schema.php:199
/app/web/db-mysql/vendor/yiisoft/db/tests/unit/CommandTest.php:1463
/app/web/db-mysql/vendor/phpunit/phpunit/phpunit:61

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

As you see, when at second time I try to launch the same test (with existing database), test will fail

rob006: What if you pass `true` as second argument of `getTableSchema()`?

rob006: Then I would remove comment above or change it to `View must be deleted before deleting the tables on which it depends` - right it creates more confusion than lack of it :P

machour: Thank you for the feedback!

tests/unit/validators/UniqueValidatorTest.php
machour: Not needed for now

hiqsol: In this case (in tests) I would prefer `factory->create` to have it shorter and more expressive.
In main framework code I would prefer not abuse DI where it is possible.
And I prefer NOT to use `Yii::createObject` and Yii helper in general.

PR #48, opened 2019-01-06
Fix tests

FabrizioCaldarelli:
This PR also solves a test fail for db-mysql (https://github.com/yiisoft/db/compare/master...FabrizioCaldarelli:fix-tests?expand=1#diff-605b1b829c7a2ab3747175bc34ff7b39)


src/ForeignKeyConstraint.php
```diff
@@ -18,21 +18,64 @@ class ForeignKeyConstraint extends Constraint
...
+     * Constructor
+     * @param string|null name (inherited from parent) the constraint name.
+     * @param string[]|null columnNames (inherited from parent) list of column names the constraint belongs to.
+     * @param string|null foreignSchemaName referenced table schema name.
+     * @param string foreignTableName referenced table name.
+     * @param string[] foreignColumnNames list of referenced table column names.
+     * @param string|null onUpdate referential action if rows in a referenced table are to be updated.
+     * @param string|null onDelete referential action if rows in a referenced table are to be deleted.
+     */
+    public function __construct(?string $name, ?iterable $columnNames, ?string $foreignSchemaName, string $foreignTableName, iterable $foreignColumnNames, ?string $onDelete, ?string $onUpdate)
```
rob006: It looks like big regression to me. Previously you have very readable key-based configuration array, now you need to remember order of 7 constructor arguments to understand what is what.

rob006:
How this is even possible that tests are passing if tests are using old signature for constructors:

https://github.com/yiisoft/db/blob/aa20542a8ec770e032f63a3872354cc1402a3be5/tests/unit/SchemaTest.php#L669-L676

src/ForeignKeyConstraint.php
FabrizioCaldarelli: I have thought these two points:
- avoid different syntax and approach with other *Constraint classes;
- how many times you use ForeignKeyConstraint directly to have to pass every parameters?

If direction is to have constructor based classes, I think that in this case we could go on.

src/Constraint.php
```diff
@@ -20,9 +20,29 @@ class Constraint extends BaseObject
...
      */
-    public $name;
+    private $name;
+
+    /**
+     * Constructor
+     * @param string|null $name the constraint name.
+     * @param string[]|null $columnNames list of column names the constraint belongs to.
+     */
+    public function __construct(?string $name, ?iterable $columnNames)
```
samdark: Does it make sense to have name/columnNames set to null here?

src/ForeignKeyConstraint.php
samdark: Why all these are null-able?

samdark: Optional parameters may go into setters.

FabrizioCaldarelli: If I put null-able parameters out of construtor parameters, the syntax to create and initialize object can be even harder, because developer should be:

1) Create the object with : `new ForeignKeyConstraint`, passing only needed parameters;
2) Set eventually optional parameters, one by one; here we have 4 null-able parameters, it  will need to write, eventualy, other 4 assigments;

FabrizioCaldarelli: And what about Constraint construtor? It has all 2 fields null-able. 

Do we want to make a constructor without parameters or anyway we use parameters in constructor?

This could be a guideline. We could create the constructor for only not null-able parameters, and others filled using setters.

What do you thing?

samdark: Thought about it more. Since the class isn't used by end users, the most straightforward way is to keep old array-based constructor.

hiqsol: This class is a good example where I would prefer to extend from `ConfigurableObject`.
https://github.com/yiisoft/yii-core/pull/136

samdark:
Merged. Thanks!

PR #55, opened 2019-01-25
Removed parent constructor call

FabrizioCaldarelli:


hiqsol:
Thank you!
As you can see unfortunately there are no tests for this class.
It would be cool add at least the most basic test for such classes: just test object construction.
It will be a starting point for better test coverage.

FabrizioCaldarelli:
I can see that and I was thinking the same thing, to start writing some tests.

I'll do soon.

PR #107, opened 2020-02-17
Enhacement DatabaseFactory for master and slave.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

- [x] Create DatabaseFactory independent, will allow to create the master and slave connections through the factory with autowired.

Example:

```php
$connection->masters[] = [
    'setDsn()'      => ['mysql:host=127.0.0.1;dbname=yiitest'],
    'setUsername()' => ['root],
    'setPassword()' => ['root'],
];

$connection->open();
```

- [x] Remove dsn from the constructor: The dsn of the constructor was removed and two setters were created for its configuration.

Example Array:

```php
$db = new Connection($cache, $logger, $profiler);

$dsn => [
     'driver' => 'mysql',
     'host'   => '127.0.0.1',
     'dbname' => 'yiitest',
],

$db->setBuildDsn($dsn);
```

Example string:

```php
$db = new Connection($cache, $logger, $profiler);

$db->setDsn('mysql:host=127.0.0.1;dbname=yiitest');
```
- [x] Change name badged README.md phpunit --> build.


xepozz:
I don't like it. Let's avoid lazy initializing via `$config = [...]`. 
I think it can be developed with `ReplicationConnectionInterface` (or smth else), that will have methods `addMaster(ConnectionInterface $connection)`, `addSlave(ConnectionInterface $connection)` and other.

terabytesoftw:
> I don't like it. Let's avoid lazy initializing via `$config = [...]`.
> I think it can be developed with `ReplicationConnectionInterface` (or smth else), that will have methods `addMaster(ConnectionInterface $connection)`, `addSlave(ConnectionInterface $connection)` and other.

In this particular case the factory avoids two things, one injecting the dependencies, and two possibility of creating dynamic connections, if I define everything in the container di that cannot be done.

samdark:
Is connection possible without DSN?

yii-extensions:
> Is connection possible without DSN?

You can create the `Connection::class` without configuring `Dsn, Username and Password`, but when you are going to open the `$connection->open()` connection, if it is not configured, throw an exception, example:

https://github.com/yiisoft/db/blob/aad933b2b33f7ae210e10c4c4fc7dea03a542a2e/src/Drivers/Connection.php#L689

terabytesoftw:
> Is connection possible without DSN?

You can create the `Connection::class` without configuring `Dsn, Username and Password`, but when you are going to open the `$connection->open()` connection, if it is not configured, throw an exception, example:

https://github.com/yiisoft/db/blob/aad933b2b33f7ae210e10c4c4fc7dea03a542a2e/src/Drivers/Connection.php#L689

samdark:
If we can't use Connection without DSN, it's better to have it as a mandatory requirement i.e. keep it in constructor.

terabytesoftw:
> If we can't use Connection without DSN, it's better to have it as a mandatory requirement i.e. keep it in constructor.

Its mandatory for `PDO::class` not for `Connection::class`, which does too many things, and these changes are to decouple that class.

samdark:
Thanks!

PR #113, opened 2020-03-01
Improve connection class.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

- [x] move phpdoc property private to setters and getters.
- [x] change visibility property public to private.

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -1297,12 +1297,12 @@ class Command
      */
     protected function logQuery(string $category): array
     {
-        if ($this->db->getEnableLogging()) {
+        if ($this->db->isLoggingEnabled()) {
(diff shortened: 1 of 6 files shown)


src/Commands/Command.php
```diff
@@ -1297,12 +1297,12 @@ public function execute(): int
      */
     protected function logQuery(string $category): array
     {
-        if ($this->db->enableLogging) {
+        if ($this->db->getEnableLogging()) {
```
samdark: Usually we call methods returned boolean flags like `isLoggingEnabled()`.

PR #123, opened 2020-03-08
Improve querys classes.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️

- [x] add getters and setters.
- [x] move phpdoc property private to setters and getters.
- [x] change visibility property public to private.
- [x] fixed tests.


src/Querys/Query.php
```diff
@@ -511,7 +474,7 @@ protected function queryScalar($selectExpression)
                 $command = $this->createCommand();
             } catch (\Exception $e) {
                 // throw it later
-            } catch (\Throwable $e) {
+            } catch (Throwable $e) {
```
rustamwin: Can we merge these two exceptions?

terabytesoftw:
> Folders names should be singular instead of plural.

Once all the classes are reviewed I will change it to singular.

src/Querys/Query.php
terabytesoftw: done.

PR #171, opened 2020-09-07
Add methods connectioninterface

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

I am adding the methods that PSALM indicates to me in the implementations, so we do not add unused methods.

### DIFF
--- a/src/Connection/ConnectionInterface.php
+++ b/src/Connection/ConnectionInterface.php
@@ -42,8 +42,8 @@ interface ConnectionInterface
      * Please refer to the [PHP manual](https://secure.php.net/manual/en/pdo.construct.php) on the format of the DSN
      * string.
      *
-     * For [SQLite](https://secure.php.net/manual/en/ref.pdo-sqlite.connection.php) you may use a [path alias](guide:concept-aliases)
-     * for specifying the database path, e.g. `sqlite:@app/data/db.sql`.
+     * For [SQLite](https://secure.php.net/manual/en/ref.pdo-sqlite.connection.php) you may use a
+     * [path alias](guide:concept-aliases) for specifying the database path, e.g. `sqlite:@app/data/db.sql`.
      *
      * {@see charset}
      */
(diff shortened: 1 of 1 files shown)


src/Connection/ConnectionInterface.php
```diff
@@ -24,10 +25,78 @@ interface ConnectionInterface
...
+     * `;charset=UTF-8` to the DSN string.
+     *
+     * The same applies for if you're using GBK or BIG5 charset with MySQL, then it's highly recommended to specify
+     * charset via {@see dsn} like `'mysql:dbname=mydatabase;host=127.0.0.1;charset=GBK;'`.
+     *
+     * @param string|null $value
+     *
+     * @return void
+     */
+    public function setCharset(?string $value): void;
```
samdark, line 91: Do we need to set it in runtime?

src/Connection/ConnectionInterface.php
```diff
@@ -24,10 +25,78 @@ interface ConnectionInterface
...
+
+    /**
+     * Whether to enable read/write splitting by using {@see setSlaves()} to read data. Note that if {@see setSlaves()}
+     * is empty, read/write splitting will NOT be enabled no matter what value this property takes.
+     *
+     * @param bool $value
+     *
+     * @return void
+     */
+    public function setEnableSlaves(bool $value): void;
```
samdark, line 101: Do we need to set it in runtime?

terabytesoftw: Error in `yiisoft/yii-db-migration`.

```php
ERROR: UndefinedInterfaceMethod - src\Migration.php:59:20 - Method Yiisoft\Db\Co nnection\ConnectionInterface::setEnableSlaves does not exist (see https://psalm. dev/181)                                                                                 $this->db->setEnableSlaves(false);
```

src/Connection/ConnectionInterface.php
terabytesoftw: Error in `yiisoft/yii-db-migration`

```php

ERROR: UndefinedInterfaceMethod - src\Service\Database\ListTablesService.php:45: 63 - Method Yiisoft\Db\Connection\ConnectionInterface::getDSN does not exist (se e https://psalm.dev/181)                                                                 $dbname = $this->getDsnAttribute('dbname', $this->db->getDSN());
```

which refers to charset.

terabytesoftw: done remove no affect static analisys.

src/Connection/ConnectionInterface.php
samdark: Yes, I know. I'm asking if we need this method in both interface and Connection class.

terabytesoftw: Only in the interface, in the connection methods of the implementation I think that we must have the methods that have change by the driver is my opinion, the rest in the abstract class.

PR #176, opened 2020-09-11
Add method isEmulateExecution() in QueryTrait::class.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

### DIFF
--- a/src/Query/QueryTrait.php
+++ b/src/Query/QueryTrait.php
@@ -457,7 +457,7 @@ trait QueryTrait
         return $this->indexBy;
     }
 
-    public function isEmulateExecution(): bool
+    public function shouldEmulateExecution(): bool
     {
         return $this->emulateExecution;
     }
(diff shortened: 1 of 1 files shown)


src/Query/QueryTrait.php
```diff
@@ -456,4 +456,9 @@ public function getIndexBy()
     {
         return $this->indexBy;
     }
+
+    public function isEmulateExecution(): bool
```
samdark, line 460:
```suggestion
    public function shouldEmulateExecution(): bool
```

samdark (changes requested):

src/Query/QueryTrait.php
terabytesoftw: done.

samdark (approved):

PR #178, opened 2020-09-16
Enhacements dataproviders.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/Data/DataProviderInterface.php
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * DataProviderInterface is the interface that must be implemented by data provider classes.
+ *
+ * Data providers are components that sort and paginate data, and provide them to widgets such as
+ * {@see Yiisoft\Yii\DataView\GridView}, {@see Yiisoft\Yii\DataView\ListView}.
+ *
+ * For more details and usage information on DataProviderInterface, see the
+ * [guide article on data providers](guide:output-data-providers).
+ */
+interface DataProviderInterface
```
samdark, line 18: How about using interfaces from `yiisoft/data` instead of having a special one?

terabytesoftw: The problem is that for now most of them are incompatible, but I can do it in another PR, it is easier to improve everything, when it is already working.

Issue #152, opened 2020-06-09
Static analysis should be corrected with psalm.


terabytesoftw:
- [ ] fix analysis static with phan.

darkdef:
Psalm level is now - 1

PR #181, opened 2020-09-25
Fixed all issues psalm, more clean code.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/db/issues/152

Check: `vendor/bin/psalm src/Query`


src/Expression/ExpressionBuilderInterface.php
```diff
@@ -16,7 +16,7 @@ interface ExpressionBuilderInterface
      * @param ExpressionInterface $expression the expression to be built.
      * @param array $params the binding parameters.
      *
-     * @return string the raw SQL that will not be additionally escaped or quoted.
+     * @return string|array the raw SQL that will not be additionally escaped or quoted.
      */
-    public function build(ExpressionInterface $expression, array &$params = []): string;
+    public function build(ExpressionInterface $expression, array &$params = []);
```
samdark, line 21: Is that possible to make `HashCondition` to always build into string so this type-hint is left as is? It make types more strict in many places.

terabytesoftw: Fixed.

src/Query/Conditions/BetweenColumnsConditionBuilder.php
```diff
@@ -41,17 +49,23 @@ public function build(ExpressionInterface $expression, array &$params = []): str
      * @param Query|ExpressionInterface|string $columnName
      * @param array $params the binding parameters.
      *
-     * @return string
+     * @throws Exception|InvalidArgumentException|InvalidConfigException|NotSupportedException
+     *
+     * @return array|string
      */
-    protected function escapeColumnName($columnName, array &$params = []): string
+    protected function escapeColumnName($columnName, array &$params = [])
```
samdark, line 56: That should always return a string or else code in `build()` will fail.

src/Query/Conditions/BetweenColumnsConditionBuilder.php
```diff
@@ -64,9 +78,11 @@ protected function escapeColumnName($columnName, array &$params = []): string
      * @param mixed $value
      * @param array $params passed by reference
      *
-     * @return string
+     * @throws Exception|InvalidArgumentException|InvalidConfigException|NotSupportedException
+     *
+     * @return array|string
      */
-    protected function createPlaceholder($value, array &$params): string
+    protected function createPlaceholder($value, array &$params)
```
samdark, line 85: That should always return a string or else code in `build()` will fail.

src/Query/Conditions/BetweenConditionBuilder.php
```diff
@@ -44,9 +52,11 @@ public function build(ExpressionInterface $expression, array &$params = []): str
      * @param mixed $value
      * @param array $params passed by reference
      *
-     * @return string
+     * @throws InvalidArgumentException|Exception|InvalidConfigException|NotSupportedException
+     *
+     * @return array|string
      */
-    protected function createPlaceholder($value, array &$params): string
+    protected function createPlaceholder($value, array &$params)
```
samdark, line 59: That should always return a string or else code in `build()` will fail.

samdark (changes requested):

src/Query/Conditions/BetweenConditionBuilder.php
terabytesoftw: fixed.

src/Query/Conditions/BetweenColumnsConditionBuilder.php
terabytesoftw: fixed.

src/Query/Conditions/BetweenColumnsConditionBuilder.php
terabytesoftw: fixed.

samdark (approved):

samdark:
👍

PR #183, opened 2020-10-10
Change the ConnectionInterface::class visibility protected to private, remove null.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Constraint/ForeignKeyConstraint.php
+++ b/src/Constraint/ForeignKeyConstraint.php
@@ -10,7 +10,9 @@ namespace Yiisoft\Db\Constraint;
 class ForeignKeyConstraint extends Constraint
 {
     private ?string $foreignSchemaName = null;
+    /** @var object|string|null $foreignTableName */
     private $foreignTableName;
+    /** @var array|string $foreignColumnNames */
     private $foreignColumnNames;
     private ?string $onUpdate = null;
     private ?string $onDelete = null;
@@ -25,7 +27,7 @@ class ForeignKeyConstraint extends Constraint
         return $this->foreignTableName;
     }
 
-    public function getForeignColumnNames(): array
+    public function getForeignColumnNames()
     {
         return $this->foreignColumnNames;
     }
@@ -43,7 +45,7 @@ class ForeignKeyConstraint extends Constraint
     /**
      * @param string|null $value referenced table schema name.
      *
-     * @return self
+     * @return $this
      */
     public function foreignSchemaName(?string $value): self
     {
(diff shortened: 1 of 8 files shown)


src/Query/QueryBuilder.php
```diff
@@ -1920,7 +1921,7 @@ public function buildWithQueries($withs, &$params): string
     }
 
     /**
-     * @return ConnectionInterface|null the database connection.
+     * @psalm-return Connection
```
samdark, line 1924: That looks suspicious.

src/Constraint/ForeignKeyConstraint.php
```diff
@@ -10,7 +10,9 @@
 class ForeignKeyConstraint extends Constraint
 {
     private ?string $foreignSchemaName = null;
+    /** @var object|string|null $foreignTableName */
```
samdark, line 13: Are you sure it should anything but the nullable string? What object could it be?

src/Constraint/ForeignKeyConstraint.php
```diff
@@ -10,7 +10,9 @@
...
     private $foreignTableName;
+    /** @var array|string $foreignColumnNames */
```
samdark, line 15: Do we really need it to be string?

src/Constraint/ForeignKeyConstraint.php
```diff
@@ -25,7 +27,7 @@ public function getForeignTableName()
         return $this->foreignTableName;
     }
 
-    public function getForeignColumnNames(): array
+    public function getForeignColumnNames()
```
samdark, line 30: Looked correct before...

src/Query/QueryBuilder.php
terabytesoftw: Not because the interface does not have all the methods, you can indicate the implementation, it is a valid psalm annotation, but it will give an error that the method does not exist in the interface.

src/Constraint/ForeignKeyConstraint.php
terabytesoftw: yes null its used in tests Mysql.

src/Constraint/ForeignKeyConstraint.php
terabytesoftw: Fixed.

src/Constraint/ForeignKeyConstraint.php
terabytesoftw: Fixed.

samdark (approved):

samdark:
Thank you.

PR #185, opened 2020-10-12
Fix key cache string.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Connection/Connection.php
+++ b/src/Connection/Connection.php
@@ -868,7 +868,7 @@ abstract class Connection implements ConnectionInterface
             /* @var $db Connection */
             $db = DatabaseFactory::createClass($config);
 
-            $key = $this->getCacheKey(__METHOD__ . '::' . $db->getDsn());
+            $key = $this->getCacheKey([__METHOD__, $db->getDsn()]);
 
             if ($this->schemaCache instanceof CacheInterface && $this->schemaCache->get($key)) {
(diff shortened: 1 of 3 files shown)


src/Connection/Connection.php
```diff
@@ -1408,4 +1407,9 @@ public function useMaster(callable $callback)
 
         return $result;
     }
+
+    private function getCacheKey(string $key): string
+    {
+        return ctype_alnum($key) && mb_strlen($key, '8bit') <= 32 ? $key : md5($key);
```
samdark, line 1413: Is it really needed?

src/Schema/Schema.php
```diff
@@ -717,14 +717,12 @@ public function getServerVersion(): string
...
-        return [
-            __CLASS__,
-            $this->db->getDsn(),
-            $this->db->getUsername(),
-            $this->getRawTableName($name),
-        ];
+        $key = __CLASS__ . ':' . $this->db->getDsn() . ':' . $this->db->getUsername() . ':' .
+            $this->getRawTableName($name);
+
+        return ctype_alnum($key) && mb_strlen($key, '8bit') <= 32 ? $key : md5($key);
```
samdark, line 725: Is it really needed?

samdark (approved):

PR #189, opened 2020-10-22
Refactoring Dsn helper class.

mj4444ru:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |

### DIFF
--- a/src/Helper/Dsn.php
+++ b/src/Helper/Dsn.php
@@ -6,9 +6,9 @@ namespace Yiisoft\Db\Helper;
 
 final class Dsn
 {
-    private ?string $dbname;
+    private string $dbname;
     private string $driver;
-    private ?string $host;
+    private string $host;
     private ?string $port;
     private array $options;
 
@@ -30,13 +30,13 @@ final class Dsn
      *
      * ```php
      * $dsn = new Dsn('mysql', '127.0.0.1', 'yiitest', '3306');
-     * $connection = new Connection($this->cache, $this->logger, $this->profiler, $dsn->getDsn());
+     * $connection = new Connection($this->cache, $this->logger, $this->profiler, (string)$dsn);
      * ```
      *
      * Will result in the DSN string `mysql:host=127.0.0.1;dbname=yiitest;port=3306`.
      */
 
-    public function getDsn(): string
+    public function __toString(): string
     {
         $dsn = "$this->driver:" . "host=$this->host" . ';' . "dbname=$this->dbname";
 
@@ -55,10 +55,5 @@ final class Dsn
         }
 
         return $dsn;
-    }
-
-    public function getDriver(): string
-    {
-        return $this->driver;
     }
 }


terabytesoftw:
Why use a magic method, if I need to get the dsn in any way, I just use getDsn ()?

mj4444ru:
> Why use a magic method, if I need to get the `dsn` in any way, i just use `getDsn()` ?

The only function of this class is to return a string. Better if it will return the result when converting itself into a string.

mj4444ru:
> Why use a magic method, if I need to get the `dsn` in any way, i just use `getDsn()` ?

https://github.com/yiisoft/db/blob/a109d57ded66018a6bcf4350f55c336c5fbd8e3d/src/Schema/ColumnSchemaBuilder.php#L240-L254

mj4444ru:
> Why use a magic method, if I need to get the `dsn` in any way, i just use `getDsn()` ?

https://github.com/yiisoft/db/blob/a109d57ded66018a6bcf4350f55c336c5fbd8e3d/src/Schema/ColumnSchemaBuilder.php#L240-L254

So here the name of the class does not correspond to its functionality.

src/Helper/Dsn.php
```diff
@@ -6,9 +6,9 @@
 
 final class Dsn
 {
-    private ?string $dbname;
+    private string $dbname;
```
samdark, line 9: It is OK not to specify database name. That's useful, for example, when creating databases.

src/Helper/Dsn.php
```diff
@@ -6,9 +6,9 @@
...
     private string $driver;
-    private ?string $host;
+    private string $host;
```
samdark, line 11: It is OK not to speficy host. For example, SQLite uses `sqlite:/opt/databases/mydb.sq3` as connection string.

src/Helper/Dsn.php
```diff
@@ -56,9 +56,4 @@ public function getDsn(): string
 
         return $dsn;
     }
-
-    public function getDriver(): string
```
samdark, line 60: Why removing `getDriver()`?

samdark (changes requested):

src/Helper/Dsn.php
mj4444ru: 1. Not used anywhere.
2. It is necessary either to make all properties available, or none.

samdark: > It is necessary either to make all properties available, or none.

Why is it necessary?

mj4444ru: > Why is it necessary?

All other things being equal, making a getter for only one property is discrimination :)

src/Helper/Dsn.php
mj4444ru: This requires more code changes.

src/Helper/Dsn.php
mj4444ru: I thought about it, but I still need to think about it.

It seems to me that we need to make the class abstract so that it does not know about any parameters other than the driver name and allow the drivers to extend this class for their databases.

How do you look at it?

samdark:
Discsussed with @mj4444ru in private. That has to be re-done conceptually a bit.

samdark:
Won't be merged in its current state.

PR #190, opened 2020-10-22
Refactoring  class ColumnSchema and added class ColumnExtendedSchema

mj4444ru:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | 

- For the `ColumnSchema.php` class, all setters have a `set` added.

- For class `ColumnSchema.php` added fields `$schemaName`, `$tableName`;

- For the `ColumnSchema.php` class, the setters for the properties field `$name` are removed, their setting has been moved to the constructor.

- Added class `ColumnExtendedSchema`:

```php
class ColumnExtendedSchema extends ColumnSchema
{
    private ?int $maxChars = null;
    private ?int $maxBytes = null;
    private ?string $charset = null;
    private ?string $check = null;

    ...
}
```

It would be nice if components like gii could request this information to better generate models. Not all drivers will be able to give this information, but it is better to make it common so that components like `gii` would not think about what driver is.

**PS:** If the PR is accepted, I will PR for other packages that used these setters.

### DIFF
--- a/src/Schema/ColumnExtendedSchema.php
+++ b/src/Schema/ColumnExtendedSchema.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Schema;
+
+/**
+ * ColumnExtendedSchema class describes the metadata of a column in a database table.
+ */
+class ColumnExtendedSchema extends ColumnSchema
+{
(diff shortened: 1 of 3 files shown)


terabytesoftw:
closed due to inactivity

PR #193, opened 2020-11-06
Make compatible mssql.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

### DIFF
--- a/src/Connection/Dsn.php
+++ b/src/Connection/Dsn.php
@@ -30,18 +30,24 @@ final class Dsn
      *
      * ```php
      * $dsn = new Dsn('mysql', '127.0.0.1', 'yiitest', '3306');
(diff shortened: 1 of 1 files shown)


src/Connection/Dsn.php
```diff
@@ -66,4 +72,26 @@ public function getDriver(): string
...
+        $dsn = "$this->driver:" . "host=$this->host" . ';' . "dbname=$this->databaseName";
+
+        if ($this->port !== null) {
+            $dsn .= ';' . "port=$this->port";
+        }
+
+        return $dsn;
+    }
+
+    private function buildDsnSqlSrv(): string
```
samdark, line 87: If there's a difference like that, it should be on each driver's side about forming a string from a set of options.

samdark (changes requested):

PR #205, opened 2020-12-02
Clean up

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/Cache/SchemaCache.php
```diff
@@ -38,14 +45,24 @@ public function getDuration(): int
 
     public function isExclude(string $value): bool
```
samdark, line 46:
```suggestion
    public function isExcluded(string $value): bool
```

src/Cache/SchemaCache.php
```diff
@@ -38,14 +45,24 @@ public function getDuration(): int
...
     }
 
     public function isEnabled(): bool
     {
         return $this->enabled;
     }
 
+    public function set($key, $value, $ttl = null, Dependency $dependency = null): bool
+    {
+        return $this->cache->set($key, $value, $ttl, $dependency);
```
samdark, line 63: It goes against `Psr\SimpleCache\CacheInterface` so `Yiisoft\Cache\CacheInterface` should be used in this class.

src/Cache/SchemaCache.php
```diff
@@ -6,6 +6,8 @@
 
 use Psr\SimpleCache\CacheInterface;
 use Yiisoft\Cache\CacheKeyNormalizer;
+use Yiisoft\Cache\Dependency\Dependency;
+use Yiisoft\Cache\Dependency\TagDependency;
 
 final class SchemaCache
```
samdark, line 12: Need docs overall about this class and its methods. Some aren't clear enough without docs.

samdark (changes requested):

src/Cache/SchemaCache.php
terabytesoftw: done.

src/Cache/SchemaCache.php
terabytesoftw: done.

src/Cache/SchemaCache.php
terabytesoftw: done,

src/Cache/QueryCache.php
```diff
@@ -26,16 +29,37 @@ public function __construct(CacheInterface $cache, CacheKeyNormalizer $keyNormal
...
+     * then the key will be returned back as it is, integers will be converted to strings. Otherwise,
+     * a normalized key is generated by serializing the given key and applying MD5 hashing.
+     *
+     * @param mixed $key The key to be normalized.
+     *
+     * @return string The normalized cache key.
+     */
     public function normalize($key): string
     {
         return $this->keyNormalizer->normalize($key);
```
samdark, line 45: We're going to break it again soon :)

samdark (approved):

PR #208, opened 2020-12-20
Validate key remove key in SchemaCache::class.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


samdark:
Will be fixed in file cache.

PR #210, opened 2021-01-15
Add ConnectionFactory

devanych:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

Provides lazy loading of the `Yiisoft\Db\Connection\ConnectionInterface` instance to prevent a circular reference to the connection when building container definitions.

### DIFF
--- a/src/Factory/ConnectionFactory.php
+++ b/src/Factory/ConnectionFactory.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Factory;
+
+use Psr\Container\ContainerInterface;
+use RuntimeException;
+use Yiisoft\Db\Connection\ConnectionInterface;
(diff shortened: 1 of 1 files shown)


terabytesoftw (approved):
Great job 👍

samdark (approved):

samdark:
One thing is that there's no dependency on DI in the package: https://github.com/yiisoft/db/blob/master/composer.json#L20

terabytesoftw:
> One thing is that there's no dependency on DI in the package: https://github.com/yiisoft/db/blob/master/composer.json#L20

We can add to the documents how to configure the factory as a service provider, without adding `yiisoft/di` to the composer.json.

devanych:
Can add `yiisoft/factory` to `require'.

PR #211, opened 2021-01-17
Add the ability to clear QueryCache.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/yii-db-migration/pull/80

### DIFF
--- a/src/Cache/QueryCache.php
+++ b/src/Cache/QueryCache.php
@@ -26,9 +26,9 @@ final class QueryCache
         $this->cache = $cache;
     }
 
-    public function clear(): void
+    public function clear(): bool
     {
-        $this->cache->clear();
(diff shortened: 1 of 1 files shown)


src/Cache/QueryCache.php
```diff
@@ -26,6 +26,11 @@ public function __construct(CacheInterface $cache)
         $this->cache = $cache;
     }
 
+    public function clear(): bool
+    {
+        return $this->cache->clear();
```
vjik, line 31: This is remove all cache in `CacheInterface` implementation, need remove only queries cache...

PR #245, opened 2021-10-04
Fix return type TableSchema::getcolumn().

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
         "ext-mbstring": "*",
         "ext-pdo": "*",
         "psr/log": "^1.1",
-        "yiisoft/active-record": "^3.0@dev",
         "yiisoft/arrays": "^1.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -23,6 +23,7 @@
         "ext-mbstring": "*",
         "ext-pdo": "*",
         "psr/log": "^1.1",
+        "yiisoft/active-record": "^3.0@dev",
```
samdark, line 26: That should not be there.

samdark (changes requested):

composer.json
terabytesoftw: done.

samdark (approved):

samdark:
👍

PR #247, opened 2021-10-06
Draft of Table helper for using this without string in complex case

darkdef:
(for simple quote of table names)

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |

### DIFF
--- a/.github/workflows/db-simple.yml
+++ b/.github/workflows/db-simple.yml
@@ -1,0 +1,79 @@
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 0 * * *'
+
+name: db-simple
+
+jobs:
+  tests:
+    name: PHP ${{ matrix.php }}-simple
+
+    env:
+      COMPOSER_ROOT_VERSION: "dev-master"
+      extensions: dom, json, pdo
+      key: cache-v1
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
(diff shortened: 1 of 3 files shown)


src/Schema/TableName.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @param ExpressionInterface|string $name
+     * @param string|null $alias
+     * @param ExpressionInterface|string|null $schema
+     */
+    public function __construct($name, ?string $alias = null, $schema = null)
+    {
+        if (!is_string($name) && !$name instanceof ExpressionInterface) {
+            throw new InvalidArgumentException(
+                'Name of the table should be a string or an instanceof ExpressionInterface.'
```
vjik, line 46:
```suggestion
                'Name of the table should be a string or an instance of ExpressionInterface.'
```

src/Schema/TableName.php
```diff
@@ -0,0 +1,91 @@
...
+            );
+        }
+
+        if ($schema !== null && !is_string($schema) && !$schema instanceof ExpressionInterface) {
```
vjik, line 50:
```suggestion
        if ($schema !== null && !is_string($schema) && !($schema instanceof ExpressionInterface)) {
```

This suggestion only for more understandable code.

PR #252, opened 2021-12-17
Replace Connection to ConnectionInterface

iamsaint:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

This fix is ​​required so that you can use QueryBuilder with databases that work without a PDO connection. Examples of such databases are Clickhouse and restream/reindexer


terabytesoftw (approved):

samdark:
@terabytesoftw is going to fix tests.

PR #270, opened 2022-02-14
More clean code.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/Cache/QueryCache.php
```diff
@@ -16,14 +16,12 @@
  */
 final class QueryCache
 {
-    private CacheInterface $cache;
     private bool $enabled = true;
     public array $info = [];
     private ?int $duration = 3600;
 
-    public function __construct(CacheInterface $cache)
+    public function __construct(private CacheInterface $cache)
```
devanych, line 23: The package supports PHP version '7.4':
https://github.com/yiisoft/db/blob/master/composer.json#L22

terabytesoftw: hi, this is a development branch, and we agree to support `PHP 8.0` at a minimum.

PR #283, opened 2022-03-01
Set error level psalm to 2.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/Command/CommandInterface.php
```diff
@@ -523,9 +523,9 @@ public function getRawSql(): string;
     /**
      * Returns the SQL statement for this command.
      *
-     * @return string|null the SQL statement to be executed.
+     * @return string the SQL statement to be executed.
      */
-    public function getSql(): ?string;
+    public function getSql(): string;
```
darkdef, line 528: Empty string and Null - is not equal. How detect not initialized Command?

src/Command/Command.php
```diff
@@ -247,7 +246,7 @@ public function bindValues(array $values): self
                 $this->params[$value->getName()] = $value;
             } elseif (is_array($value)) { // TODO: Drop in Yii 2.1
                 $this->params[$name] = new Param($name, ...$value);
-            } elseif ($value instanceof PdoValue) {
+            } elseif ($value instanceof PdoValue && is_int($value->getType())) {
```
darkdef, line 249: This incorrect.
https://www.php.net/manual/en/pdo.constants.php
// public const PARAM_NULL = 0;

src/Query/Conditions/BetweenColumnsCondition.php
```diff
@@ -36,7 +36,7 @@
  */
 class BetweenColumnsCondition implements BetweenColumnsConditionInterface
 {
-    public function __construct(
+    final public function __construct(
```
darkdef, line 39: Why? Class not final, how about inheritance?

src/Query/Conditions/Interface/BetweenConditionInterface.php
```diff
@@ -4,15 +4,14 @@
 
 namespace Yiisoft\Db\Query\Conditions\Interface;
 
-use Yiisoft\Db\Expression\ExpressionInterface;
+use Yiisoft\Db\Expression\Expression;
```
darkdef, line 7: are sure not ExpressionInterface?

src/Query/Conditions/Interface/LikeConditionInterface.php
```diff
@@ -5,30 +5,30 @@
...
 
 interface LikeConditionInterface extends ConditionInterface
 {
     /**
-     * @psalm-return string|string[]|ExpressionInterface The column name. If it is an array, a composite `IN` condition
-     * will be generated.
+     * @psalm-return string|Expression The column name.
      */
-    public function getColumn(): string|array|ExpressionInterface;
+    public function getColumn(): string|Expression;
```
darkdef, line 17: ExpressionInterface?

terabytesoftw: The interface is only used to check that it is used but it does not have any methods, because the conditions have different methods, for example only `Expression:class` has to_string().

src/Query/Conditions/Interface/BetweenConditionInterface.php
terabytesoftw: The interface is only used to check that it is used but it does not have any methods, because the conditions have different methods, for example only Expression:class has to_string().

src/Query/Conditions/BetweenColumnsCondition.php
terabytesoftw: The psalm marks a constructor inconsistency error and recommends making it final.

PR #288, opened 2022-03-19
Split `Query::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/Query/QueryConditionsInterface.php
```diff
@@ -0,0 +1,446 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Query;
+
+use Yiisoft\Db\Exception\NotSupportedException;
+use Yiisoft\Db\Expression\ExpressionInterface;
+
+interface QueryConditionsInterface
```
darkdef, line 10: QuryConditions - incorrect name for select, from union and etc.
QueryBase ?
And how about splitting to other folders with QueryBuilder's classes?

src/Query/QueryHelper.php
```diff
@@ -0,0 +1,243 @@
...
+use function array_unshift;
+use function is_string;
+use function preg_match;
+use function preg_split;
+use function str_contains;
+use function strcasecmp;
+use function strtoupper;
+use function trim;
+
+final class QueryHelper
```
darkdef, line 23: need split tests

src/Query/QueryConditionsInterface.php
terabytesoftw: rename for `QueryParts::class`.

src/Query/QueryHelper.php
terabytesoftw: done.

PR #289, opened 2022-03-23
Ability to disable auto slave select for all read queries

sartor:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

New boolean flag for `Connection`:
Ability to disable auto slave select for all read queries. Slave connection still can be used  with `$db->getSlave()` method

### DIFF
--- a/src/Connection/Connection.php
+++ b/src/Connection/Connection.php
@@ -176,6 +176,7 @@ abstract class Connection implements ConnectionInterface
     private bool $enableSavepoint = true;
     private int $serverRetryInterval = 600;
     private bool $enableSlaves = true;
+    private bool $enableAutoSlaveForReadQueries = true;
     private array $slaves = [];
     private array $masters = [];
     private bool $shuffleMasters = true;
@@ -370,6 +371,11 @@ abstract class Connection implements ConnectionInterface
     public function areSlavesEnabled(): bool
     {
         return $this->enableSlaves;
+    }
+
+    public function isAutoSlaveForReadQueriesEnabled(): bool
+    {
+        return $this->enableAutoSlaveForReadQueries;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Connection/Connection.php
```diff
@@ -372,6 +373,11 @@ public function areSlavesEnabled(): bool
         return $this->enableSlaves;
     }
 
+    public function isAutoSlaveForReadQueriesEnabled(): bool
```
darkdef, line 376: Please add method to interface

terabytesoftw:
Please add tests in `TestConnectionTrait.php`

Thks,

src/Connection/Connection.php
sartor: There are many another methods, that used in `Command` class, but not added to `ConnectionInterface`
Is this made by design or by mistake?

terabytesoftw:
> > Please add tests in `TestConnectionTrait.php`
> 
> There is no way now to test method in class, PDO instance selection is hardcoded here. I can only write test for setter and getter in , but I think it useless. Any ideas?`prepare``Command``Connection`

It's not ok to just add tests for setter and getter, we are developing the dev branch, where all this will be possible, thanks.

src/Connection/ConnectionInterface.php
```diff
@@ -69,6 +69,15 @@ public function getServerVersion(): string;
...
+    /**
+     * Whether to enable auto recogintion of read queries and use slave (if enabled) for execute
+     *
+     * @return bool:
+     *  true -> (default) use slave for read queries, master for other
+     *  false -> always use master connection (for reads and writes). Slave still can be used via $this->getSlave()
+     */
+    public function isAutoSlaveForReadQueriesEnabled(): bool;
+
     /**
```
terabytesoftw, line 81: /**
     * Whether to enable auto recogintion of read queries and use slave (if enabled) for execute.
     *
     * @return bool For default `true` use slave for read queries, `false` use master master connection (for reads and
     * writes). Slave still can be used via $this->getSlave().
     */
    public function isAutoSlaveForReadQueriesEnabled(): **bool;
```

sartor:
Must be merged with https://github.com/yiisoft/db-redis/pull/39 because of interface change

terabytesoftw (approved):

vjik (approved):

PR #294, opened 2022-03-29
Set error level to `1` psalm - part4.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Connection/Connection.php
+++ b/src/Connection/Connection.php
@@ -63,8 +63,10 @@ abstract class Connection implements ConnectionInterface
         $this->queryCache->setInfo(
             [$duration ?? $this->queryCache->getDuration(), $dependency]
         );
+        /** @var mixed */
         $result = $callable($this);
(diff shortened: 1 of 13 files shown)


src/Connection/Connection.php
```diff
@@ -221,9 +228,9 @@ public function useMaster(callable $callback): mixed
      *
      * @param array $pool The list of connection configurations in the server pool.
      *
-     * @return static|null The opened DB connection, or `null` if no server is available.
+     * @return ConnectionPDOInterface|null The opened DB connection, or `null` if no server is available.
      */
-    protected function openFromPool(array $pool): ?self
+    protected function openFromPool(array $pool): ?ConnectionPDOInterface
```
darkdef, line 233: may be ConnectionInterface?

PR #298, opened 2022-04-03
Fix for query* functions and tests

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Command/Command.php
+++ b/src/Command/Command.php
@@ -467,19 +467,10 @@ abstract class Command implements CommandInterface
         return $this->queryInternal((int) static::QUERY_MODE_CURSOR);
     }
 
-    /**
-     * @psalm-suppress MixedReturnStatement
-     * @psalm-suppress MixedInferredReturnType
-     */
     public function queryAll(): array
     {
-        return $this->queryInternal((int) static::QUERY_MODE_ALL);
-    }
-
-    public function queryColumn(): array
-    {
         /** @psalm-var array<array-key, array<mixed>>|null */
-        $results = $this->queryInternal((int) static::QUERY_MODE_COLUMN);
+        $results = $this->queryInternal((int) static::QUERY_MODE_ALL);
 
         if (is_array($results)) {
             return $results;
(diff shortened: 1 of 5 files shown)


src/Command/Command.php
```diff
@@ -467,30 +467,40 @@ public function query(): DataReader
...
-     * @psalm-suppress MixedReturnStatement
-     * @psalm-suppress MixedInferredReturnType
-     */
     public function queryAll(): array
     {
-        return $this->queryInternal((int) static::QUERY_MODE_ALL);
+        /** @psalm-var array<array-key, array<mixed>>|null */
+        $results = $this->queryInternal((int) static::QUERY_MODE_ALL);
+
+        if (is_array($results)) {
```
terabytesoftw, line 475:
```suggestion
    return match (is_array($results) {
        true => $results,
        false => [],
    };
```

src/Command/Command.php
```diff
@@ -467,30 +467,40 @@ public function query(): DataReader
...
     }
 
-    public function queryColumn(): array
+    public function queryColumn(): array|false
     {
         /** @psalm-var array<array-key, array<mixed>>|null */
         $results = $this->queryInternal((int) static::QUERY_MODE_COLUMN);
 
-        if (is_array($results)) {
+        if (is_array($results) && count($results)) {
```
terabytesoftw, line 487:
```suggestion
    return match (is_array($results) && count($results)) {
        true => $results,
        false => false,
    };
```

src/Command/Command.php
```diff
@@ -467,30 +467,40 @@ public function query(): DataReader
...
     }
 
-    public function queryOne(): mixed
+    public function queryOne(): array|false
     {
-        return $this->queryInternal((int) static::QUERY_MODE_ROW);
+        /** @psalm-var array<array-key, array<mixed>>|null */
+        $results = $this->queryInternal((int) static::QUERY_MODE_ROW);
+
+        if (is_array($results) && count($results)) {
```
terabytesoftw, line 499:
```suggestion
    return match (is_array($results) && count($results)) {
        true => $results,
        false => false,
    };
```

src/Query/Query.php
```diff
@@ -285,7 +285,11 @@ public function column(): array
         }
 
         if ($this->indexBy === null) {
-            return $this->createCommand()->queryColumn();
+            $result = $this->createCommand()->queryColumn();
+            if (is_array($result)) {
```
terabytesoftw, line 289:
```suggestion
    return match (is_array($result)) {
        true => $result,
        false => [],
    };
```

terabytesoftw (changes requested):

src/Command/Command.php
darkdef: It's not so readable

terabytesoftw (approved):

PR #322, opened 2022-06-02
Separate chain calls

sankaest:
| Q             | A
| ------------- | ---
| Readability fix  | according to https://github.com/yiisoft/docs/issues/155, chain calls must be put on separate lines.
----
Found some more.


darkdef:
according to https://github.com/yiisoft/docs/issues/155, chain calls must be put on separate lines - is not applicable here.

Issue #17, opened 2018-11-02
Query::one() should return null instead of false, when no results


brandonkelly:
`yii\db\Query::one()` returns `false` if there are no results. In a rare break from consistency, `yii\db\ActiveQuery::one()` returns `null` instead.

I can’t think of any reason why they would need to return different things, so this is probably just an oversight.

Of the two, `null` is probably the better option, as it’s more semantic and would allow the result could be used in conjunction with a `??` operator, e.g 

```php
$foo = $query->one() ?? 'default';
```

So `Query::one()` is the method that should change.

PR #343, opened 2022-07-07
Change return type for query one and query column

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #17 

Adopt PRs:
- https://github.com/yiisoft/active-record/pull/190
- https://github.com/yiisoft/db-mssql/pull/108
- https://github.com/yiisoft/db-mysql/pull/137
- https://github.com/yiisoft/db-oracle/pull/85
- https://github.com/yiisoft/db-pgsql/pull/140
- https://github.com/yiisoft/db-sqlite/pull/118

### DIFF
--- a/src/Command/Command.php
+++ b/src/Command/Command.php
@@ -471,28 +471,20 @@ abstract class Command implements CommandInterface
         return [];
     }
 
-    public function queryColumn(): array|false
+    public function queryColumn(): array
     {
-        /** @psalm-var array<array-key, array<mixed>>|null */
-        $results = $this->queryInternal((int) static::QUERY_MODE_COLUMN);
+        /** @psalm-var mixed */
+        $results = $this->queryInternal(self::QUERY_MODE_COLUMN);
 
-        if (is_array($results) && count($results)) {
-            return $results;
-        }
-
-        return false;
+        return is_array($results) ? $results : [];
     }
 
-    public function queryOne(): array|false
+    public function queryOne(): ?array
     {
-        /** @psalm-var array<array-key, array<mixed>>|null */
-        $results = $this->queryInternal((int) static::QUERY_MODE_ROW);
+        /** @psalm-var mixed */
+        $results = $this->queryInternal(self::QUERY_MODE_ROW);
 
-        if (is_array($results) && count($results)) {
-            return $results;
-        }
-
-        return false;
+        return is_array($results) ? $results : null;
     }
 
     /**
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -608,7 +608,7 @@ interface CommandInterface
      *
      * @throws Exception|Throwable Execution failed.
      *
-     * @return array The first column of the query result.
+     * @return array The first column of the query result. Empty array is returned if the query results in nothing.
      */
     public function queryColumn(): array;
 
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -608,10 +608,9 @@ interface CommandInterface
      *
      * @throws Exception|Throwable Execution failed.
      *
-     * @return array|false The first column of the query result.
-     * False is returned if the query results in nothing.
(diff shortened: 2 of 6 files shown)


src/Command/Command.php
```diff
@@ -471,28 +471,20 @@ public function queryAll(): array
...
-        $results = $this->queryInternal((int) static::QUERY_MODE_ROW);
-
-        if (is_array($results) && count($results)) {
-            return $results;
-        }
+        /** @psalm-var mixed */
+        $results = $this->queryInternal(self::QUERY_MODE_ROW);
 
-        return false;
+        return is_array($results) ? $results : null;
```
darkdef, line 487: 1. You have error in code - some drivers(pdo and not pdo) may return empty array or null if not found rows (for example - https://www.php.net/manual/ru/mysqli-result.fetch-row.php). 
2. I think current behaviour correct - array or false.
Null it's undefined - but we have "not found data". Not "undefined"

darkdef: How about changes in \Yiisoft\Db\Pgsql\CommandPDO::insertEx (and other usages of queryIne) to behaviour array|false ?
```
        /** @var mixed */
        $queryOne = $this->queryOne();

        return is_array($queryOne) ? $queryOne : false;
```
to
```
        return $this->queryOne();
```

darkdef (changes requested):

src/Command/CommandInterface.php
```diff
@@ -608,10 +608,9 @@ public function queryAll(): array;
      *
      * @throws Exception|Throwable Execution failed.
      *
-     * @return array|false The first column of the query result.
-     * False is returned if the query results in nothing.
+     * @return array The first column of the query result.
```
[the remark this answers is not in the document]
vjik: Yes. Add this to phpdoc?

src/Command/Command.php
```diff
@@ -471,28 +471,20 @@ public function queryAll(): array
         return [];
     }
 
-    public function queryColumn(): array|false
+    public function queryColumn(): array
```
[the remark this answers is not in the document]
vjik: `![] === true` 

This code will be work fine. But it can be simplified and immediately return the result (`return $this->db->createCommand(...)->queryColumn();`)

src/Command/Command.php
vjik: Null is "without value" also. In PHP wide use return null when something not found.

src/Command/CommandInterface.php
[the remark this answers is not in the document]
vjik: Done.

src/Query/Query.php
```diff
@@ -550,7 +550,7 @@ public function offset(Expression|int|null $offset): QueryInterface
         return $this;
     }
 
-    public function one(): ?array
+    public function one(): array|object|null
```
terabytesoftw, line 553: You can add tests where each return of the method is checked, thanks.

src/Command/Command.php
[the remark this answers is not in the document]
vjik: Add adopt PRs (see issue description)

samdark (approved):

src/Query/Query.php
vjik: array and null tested [here](https://github.com/yiisoft/db/blob/7c4679af22b044567e5e35bb3a04893dac5298fd/src/TestSupport/TestQueryTrait.php#L502)

object need for active-record and tested in `yiisoft/active-record` package

terabytesoftw: Returning in an object method because another package needs it is an error, it is conditioning the package to a dependency that it does not have, db does not depend on AR, in addition to adding a nonexistent return.

darkdef: please change return type to ?array
in ActiveRecord will be changed interface

PR #391, opened 2022-11-21
Fix 8: Improve the exception message.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Schema/TableSchema.php
+++ b/src/Schema/TableSchema.php
@@ -208,6 +208,6 @@ abstract class TableSchema implements TableSchemaInterface
 
     public function compositeFK(int $id, string $from, string $to): void
     {
-        throw new NotSupportedException('Composite foreign key not supported');
+        throw new NotSupportedException(static::class . ' does not support composite FK.');
     }
 }


samdark (approved):

src/Schema/TableSchema.php
```diff
@@ -208,6 +208,6 @@ public function foreignKey(string|int $id, array $to): void
 
     public function compositeFK(int $id, string $from, string $to): void
     {
-        throw new NotSupportedException('Composite foreign key not supported');
+        throw new NotSupportedException(static::class . ' does not support composite FK.');
```
samdark, line 211:
```suggestion
        throw new NotSupportedException(self::class . ' does not support composite FK.');
```

samdark: Because you expect concrete TableSchema here, not this exact class.

samdark (changes requested):

src/Schema/TableSchema.php
samdark: Oh, sorry. You're right.

samdark (approved):

darkdef (approved):

PR #425, opened 2022-12-20
Remove `SchemaBuilderTrait::class` from `yiisoft/db`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/.github/workflows/db-pgsql.yml
+++ b/.github/workflows/db-pgsql.yml
@@ -26,7 +26,7 @@ jobs:
     name: PHP ${{ matrix.php }}-pgsql-${{ matrix.pgsql }}
 
     env:
-      COMPOSER_ROOT_VERSION: dev-remove-schema-builder-trait
+      COMPOSER_ROOT_VERSION: dev-master
       extensions: pdo, pdo_pgsql
 
     runs-on: ${{ matrix.os }}


darkdef (approved):

.github/workflows/db-pgsql.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     name: PHP ${{ matrix.php }}-pgsql-${{ matrix.pgsql }}
 
     env:
-      COMPOSER_ROOT_VERSION: dev-master
+      COMPOSER_ROOT_VERSION: dev-remove-schema-builder-trait
```
samdark, line 29: Should be adjusted before merge.

src/Migration/AbstractMigrationBuilder.php
```diff
@@ -0,0 +1,339 @@
...
+ *         'updated_at' => $this->datetime(),
+ *     ],
+ * );
+ * ```
+ */
+abstract class AbstractMigrationBuilder
+{
+    public function __construct(private SchemaInterface $schema)
+    {
+    }
```
samdark, line 39: I think it's easier to read if we'll format it like the following:

```suggestion
    public function __construct(
        private SchemaInterface $schema
    )
    {
    }
```

samdark (approved):

src/Migration/AbstractMigrationBuilder.php
darkdef: I think: not for short lines

PR #440, opened 2022-12-28
Fix phpdoc, add more tests.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/.github/workflows/db-mssql.yml
+++ b/.github/workflows/db-mssql.yml
@@ -26,7 +26,7 @@ jobs:
     name: PHP ${{ matrix.php }}-mssql-${{ matrix.mssql }}
 
     env:
-      COMPOSER_ROOT_VERSION: dev-add-more-tests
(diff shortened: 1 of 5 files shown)


darkdef (approved):

tests/Common/CommonQueryTest.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Ensure no ambiguous column error occurs on indexBy with JOIN.
+     *
+     * @link https://github.com/yiisoft/yii2/issues/13859
+     */
+    public function testAmbiguousColumnIndexBy()
+    {
+        $db = $this->getConnection(true);
+
+        $selectExpression = match ($db->getName()) {
```
darkdef, line 21: How about move this code to db-mysql, db-oracle, etc?

PR #461, opened 2023-01-06
Refactoring insertWithReturningPks

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


src/Command/AbstractCommand.php
```diff
@@ -365,6 +365,20 @@ public function insert(string $table, QueryInterface|array $columns): static
         return $this->setSql($sql)->bindValues($params);
     }
 
+    public function insertWithReturningPks(string $table, array $columns): bool|array
```
rustamwin, line 368:
```suggestion
    public function insertWithReturningPks(string $table, array $columns): false|array
```

PR #467, opened 2023-01-09
Use prefix `Abstract` in class in `src/Schema`, fix phpdoc.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Schema/TableSchemaInterface.php
+++ b/src/Schema/TableSchemaInterface.php
@@ -193,7 +193,7 @@ interface TableSchemaInterface
     /**
      * Set one foreignKey by index.
      *
-     * @param string|int $id The index of foreign key.
+     * @param int|string $id The index of foreign key.
      */
     public function foreignKey(string|int $id, array $to): void;
 


src/Schema/AbstractColumnSchema.php
```diff
@@ -270,8 +270,8 @@ protected function typecast(mixed $value): mixed
         }
 
         switch ($this->phpType) {
-            case Schema::PHP_TYPE_RESOURCE:
-            case Schema::PHP_TYPE_STRING:
+            case AbstractSchema::PHP_TYPE_RESOURCE:
```
darkdef, line 273: How about move this constants to SchemaInterface, and use SchemaInterface::PHP_TYPE_STRING etc

src/Schema/AbstractColumnSchemaBuilder.php
```diff
@@ -50,27 +50,27 @@ class ColumnSchemaBuilder implements Stringable
...
-        Schema::TYPE_DOUBLE => self::CATEGORY_NUMERIC,
-        Schema::TYPE_DECIMAL => self::CATEGORY_NUMERIC,
-        Schema::TYPE_DATETIME => self::CATEGORY_TIME,
-        Schema::TYPE_TIMESTAMP => self::CATEGORY_TIME,
-        Schema::TYPE_TIME => self::CATEGORY_TIME,
-        Schema::TYPE_DATE => self::CATEGORY_TIME,
-        Schema::TYPE_BINARY => self::CATEGORY_OTHER,
-        Schema::TYPE_BOOLEAN => self::CATEGORY_NUMERIC,
-        Schema::TYPE_MONEY => self::CATEGORY_NUMERIC,
+        AbstractSchema::TYPE_PK => self::CATEGORY_PK,
```
darkdef, line 53: And this constants

src/Schema/AbstractColumnSchemaBuilder.php
```diff
@@ -251,7 +249,7 @@ protected function buildLengthString(): string
             $this->length = implode(',', $this->length);
         }
 
-        return "({$this->length})";
+        return "($this->length)";
```
darkdef, line 252: Maybe as is? Bad for reading

darkdef (changes requested):

PR #562, opened 2023-03-02
Docs part 2.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/docs/en/getting-started.md
+++ b/docs/en/getting-started.md
@@ -85,8 +85,8 @@ Once you have a database connection instance, you can execute a **SQL** query by
 
 When writing database-agnostic code, properly quoting table and column names is often a headache because different databases have different name quoting rules. To overcome this problem, you may use the following quoting syntax introduced by [Yii DB](https://github.com/yiisoft/db):
 
-- `[[column name]]`: enclose a column name to be quoted in double square brackets.
-- `{{table name}}`: enclose a table name to be quoted in double curly brackets.
+- `[[column name]]`: enclose a *column name* to be quoted in *double square brackets*.
+- `{{%table name}}`: enclose a *table name* to be quoted in *double curly brackets*, and the percentage character `%` will be replaced with the *table prefix*.
 
 [Yii DB](https://github.com/yiisoft/db) will automatically convert such constructs into the corresponding quoted column or table names using the DBMS specific syntax.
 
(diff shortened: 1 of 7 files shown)


samdark (approved):

docs/en/getting-started.md
```diff
@@ -79,3 +79,31 @@ Once you have a database connection instance, you can execute a **SQL** query by
...
 2. [Bind parameters](/docs/en/queries/bind-parameters.md)
 3. [Call one of the SQL execute method to execute the command](/docs/en/queries/execute-command.md)
+
+
+## Quoting Table and Column Names
+
+When writing database-agnostic code, properly quoting table and column names is often a headache because different databases have different name quoting rules. To overcome this problem, you may use the following quoting syntax introduced by [Yii DB](https://github.com/yiisoft/db):
+
+- `[[column name]]`: enclose a column name to be quoted in double square brackets.
+- `{{table name}}`: enclose a table name to be quoted in double curly brackets.
```
darkdef, line 89: {{%table_name}}  - need recommended add percent before name for use prefix of tables

docs/en/query-builder.md
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+use Yiisoft\Db\Connection\ConnectionInterface;
+use Yiisoft\Db\Query\Query;
+
+/** @var ConnectionInterface $db */
+
+$rows = (new Query($db))
+    ->select(['id', 'email'])
+    ->from('user')
```
darkdef, line 24: from('{{%user}}')

and need add example with correct work with alias
['alias' => '{{%table_name}}']

docs/en/query/from.md
```diff
@@ -0,0 +1,28 @@
...
+// SELECT * FROM `user`
+$query->from('user');
+```
+
+You can specify the **table(s)** being selected from in either a string or an array. The table names may contain **schema prefixes and/or table aliases**, like you do when writing **raw SQL statements**.
+
+For example.
+
+```php
+$query->from(['public.user u', 'public.post p']);
```
darkdef, line 17: need add "we not recommed this method - use array format" -

docs/en/getting-started.md
darkdef: and need use {{%table_name}} in all documentations. It's recommended example of usage

PR #565, opened 2023-03-03
Fix phpdoc `Command` classes and `SchemaCache::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Command/AbstractCommand.php
+++ b/src/Command/AbstractCommand.php
@@ -46,7 +46,7 @@ use function strncmp;
  * $users = $connectionInterface->createCommand('SELECT * FROM user')->queryAll();
  * ```
  *
- * Supports SQL statement preparation and parameter binding.
+ * Abstract command supports SQL statement preparation and parameter binding.
  *
  * Call {@see bindValue()} to bind a value to a SQL parameter;
  * Call {@see bindParam()} to bind a PHP variable to a SQL parameter.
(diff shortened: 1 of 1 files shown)


darkdef (approved):

src/Command/AbstractCommand.php
```diff
@@ -48,15 +46,15 @@
  * $users = $connectionInterface->createCommand('SELECT * FROM user')->queryAll();
  * ```
  *
- * Command supports SQL statement preparation and parameter binding.
+ * Supports SQL statement preparation and parameter binding.
```
samdark, line 49: I think it was OK before. Here we don't have the context anymore. Last time we mentioned an object it was SQL statement so it becomes confusing.

src/Command/AbstractCommand.php
```diff
@@ -48,15 +46,15 @@
...
  *
  * Call {@see bindValue()} to bind a value to a SQL parameter;
  * Call {@see bindParam()} to bind a PHP variable to a SQL parameter.
  *
  * When binding a parameter, the SQL statement is automatically prepared. You may also call {@see prepare()} explicitly
  * to prepare a SQL statement.
  *
- * Command also supports building SQL statements by providing methods such as {@see insert()}, {@see update()}, etc.
+ * Also supports building SQL statements by providing methods such as {@see insert()}, {@see update()}, etc.
```
samdark, line 57: Same as https://github.com/yiisoft/db/pull/565/files#r1124816432

PR #568, opened 2023-03-04
Add contexts for profiler

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️


src/Profiler/ContextInterface.php
```diff
@@ -0,0 +1,19 @@
...
+interface ContextInterface
+{
+    /**
+     * @return string Type of the context
+     */
+    public function getType(): string;
+
+    public function setException(Throwable $e): static;
+
+    public function __toArray(): array;
```
xepozz, line 18: Why with `__` prefix?

src/Profiler/ContextInterface.php
```diff
@@ -0,0 +1,19 @@
...
+use Throwable;
+
+interface ContextInterface
+{
+    /**
+     * @return string Type of the context
+     */
+    public function getType(): string;
+
+    public function setException(Throwable $e): static;
```
xepozz, line 16: What the reason to have it in the interface?

xepozz (changes requested):

src/Profiler/ContextInterface.php
darkdef: fixed

xepozz (approved):

PR #582, opened 2023-03-07
Fix phpdoc `Query` classes, and add property protected `with` in `Query` class.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/Query/Query.php
```diff
@@ -82,6 +85,7 @@ class Query implements QueryInterface
     protected ExpressionInterface|int|null $limit = null;
     protected ExpressionInterface|int|null $offset = null;
     protected array|string|ExpressionInterface|null $where = null;
+    protected array $with = [];
```
samdark, line 88: Doesn't look like a docs change.

samdark (approved):

Issue #598, opened 2023-03-18
`ColumnSchemaInterface' does not guarantee the availability of data


vjik:
For example, `getName()` by interface can return empty string. How can there be a column without a name?

samdark:
https://github.com/yiisoft/db/blob/master/src/Schema/ColumnSchemaInterface.php#L206

samdark:
"By default, it returns an empty string." should be removed from phpdoc.

As for implementations, we may have __construct(string $name) in ColumnSchema. But that part is optional since users will hardly ever have to create it manually.

PR #600, opened 2023-03-18
Remove unncessary assumptions on default values from phpdoc of ColumnSchemaInterface, make name mandatory

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #598

### DIFF
--- a/src/Schema/ColumnSchemaInterface.php
+++ b/src/Schema/ColumnSchemaInterface.php
@@ -202,7 +202,6 @@ interface ColumnSchemaInterface
 
     /**
      * @return string The name of the column.
-     *
      * @psalm-return non-empty-string
      */
     public function getName(): string;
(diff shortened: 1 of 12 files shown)


darkdef (approved):

src/Schema/ColumnSchemaInterface.php
```diff
@@ -221,6 +219,8 @@ public function getPrecision(): int|null;
      * @return string|null The phpType of the column. `null` if no phpType has been defined.
      * By default, it returns `null`.
      *
+     * @psalm-return non-empty-string
```
vjik, line 222:
```suggestion
     * @psalm-return non-empty-string|null
```

samdark:
Merged into https://github.com/yiisoft/db/pull/601

PR #623, opened 2023-03-27
Fix exception `SchemaCache::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Exception/PsrInvalidArgumentException.php
+++ b/src/Exception/PsrInvalidArgumentException.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Yiisoft\Db\Exception;
 
-use Yiisoft\Db\Exception\Exception;
-
 /**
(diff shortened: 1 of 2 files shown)


src/Exception/PsrInvalidArgumentException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Exception;
+
+/**
+ * Represents an exception that's caused by invalid operations of cache.
+ */
+final class PsrInvalidArgumentException extends Exception implements \Psr\SimpleCache\InvalidArgumentException
```
samdark, line 10: Why do we need this proxy class?

darkdef (approved):

samdark (approved):

PR #624, opened 2023-03-29
Add phpdoc constans `AbstractColumn::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Schema/Builder/AbstractColumn.php
+++ b/src/Schema/Builder/AbstractColumn.php
@@ -30,33 +30,33 @@ use function strtr;
 abstract class AbstractColumn implements ColumnInterface
 {
     /**
-     * Define the type of the column as a primary key.
+     * Allows you to group and define the abstract column type as primary key.
      */
-    public const CATEGORY_PK = 'pk';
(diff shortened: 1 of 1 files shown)


src/Schema/Builder/AbstractColumn.php
```diff
@@ -30,16 +30,32 @@
 abstract class AbstractColumn implements ColumnInterface
 {
     /**
-     * Internally used constants representing categories that abstract column types fall under.
-     *
-     * {@see $categoryMap} For mappings of abstract column types to category.
+     * Define the type of the column as a primary key.
      */
     public const CATEGORY_PK = 'pk';
```
samdark, line 35: Hmm... so these are types? Maybe it worth changing the constants name from `CATEGORY_` to `TYPE_`?

samdark: Are you sure that the phpdoc is correct then? It doesn't mention anything about this and states that it's type.

samdark (approved):

PR #646, opened 2023-04-04
PHP Docs Improvements

cebe:
Also changed `Query::addParam()` if structure to return early.

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -133,20 +133,22 @@ class Query implements QueryInterface
 
     public function addParams(array $params): static
     {
-        if (!empty($params)) {
-            if (empty($this->params)) {
-                $this->params = $params;
-            } else {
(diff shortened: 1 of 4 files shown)


src/Query/Query.php
```diff
@@ -133,20 +133,22 @@ public function addOrderBy(array|string|ExpressionInterface $columns): static
...
+        } else {
+            /**
+             * @psalm-var array $params
+             * @psalm-var mixed $value
+             */
+            foreach ($params as $name => $value) {
+                if (is_int($name)) {
+                    $this->params[] = $value;
+                } else {
+                    $this->params[$name] = $value;
```
cebe, line 151: does this reduce code coverage? :-D
if so, there are some tests missing :)

samdark (approved):

PR #656, opened 2023-04-05
Better naming classes with `PDO` - part 2.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/Driver/Pdo/AbstractCommand.php
```diff
@@ -24,7 +23,7 @@
  *
  * It also provides methods for binding parameter values and retrieving query results.
  */
-abstract class AbstractCommandPDO extends AbstractCommand implements CommandPDOInterface
+abstract class AbstractCommand extends \Yiisoft\Db\Command\AbstractCommand implements CommandInterface
```
darkdef, line 26: I'm not right with naming
Better 
prefix: Abstract
type: Pdo
class Command


AbstractPdoCommand
AbstractPdoConnection
etc...

in subpackages 
PdoCommand
PdoConnection

and from other drivers (for example odbc)
OdbcCommand
OdbcConnection

samdark: Not doing the renaming is fine for me as well as doing it.

samdark: @terabytesoftw, @darkdef decide it please.

PR #667, opened 2023-04-06
Various corrections before release, part 2

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/docs/en/command/ddl.md
+++ b/docs/en/command/ddl.md
@@ -227,13 +227,13 @@ use Yiisoft\Db\Connection\ConnectionInterface;
 
 /** @var ConnectionInterface $db */
 $db->createCommand()->addForeignKey(
-    '{{%customer}}',
-    'fk-customer-profile_id',
-    'profile_id',
-    '{{%profile}}',
-    'id',
-    'CASCADE',
-    'CASCADE'
+    table: '{{%customer}}',
+    name: 'fk-customer-profile_id',
+    columns: 'profile_id',
+    refTable: '{{%profile}}',
+    refColumns: 'id',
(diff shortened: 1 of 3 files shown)


docs/en/command/ddl.md
```diff
@@ -227,13 +227,13 @@ use Yiisoft\Db\Connection\ConnectionInterface;
...
 /** @var ConnectionInterface $db */
 $db->createCommand()->addForeignKey(
-    '{{%customer}}',
-    'fk-customer-profile_id',
-    'profile_id',
-    '{{%profile}}',
-    'id',
-    'CASCADE',
-    'CASCADE'
+    table: '{{%customer}}',
```
darkdef, line 230: o'm not like named params

samdark: In this case we can provide it as is w/o named arguments.

arogachev: @darkdef Why? In case of a bigger amount of arguments / when it's hard to tell by values what the argument names are, it's useful and convenient.

@samdark Surely we can, but it improves readability in this case, no need to look for method signature / description.

They also can be applied partially.

arogachev: In this particular example, it's hard to differentiate between `table` / `refTable`, `columns` / `refColumns` and `delete` / `update`.

samdark: Reverted this part. Will decide if we want it after release.

samdark (approved):

PR #675, opened 2023-04-07
Fixes

samdark:
Assorted phpdoc and naming fixes. If something isn't correct please comment and I'll remove changes.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -138,8 +138,8 @@ interface CommandInterface
     /**
      * Creates an SQL command for changing the definition of a column.
      *
-     * @param string $table The table whose column is to be changed.
-     * @param string $column The name of the column to be changed.
+     * @param string $table The table whose column is to change.
+     * @param string $column The name of the column to change.
      * @param string $type The column type. {@see QueryBuilder::getColumnType()} will be called to convert the give
      * column type to the physical one. For example, `string` will be converted as `varchar(255)`, and `string not null`
      * becomes `varchar(255) not null`.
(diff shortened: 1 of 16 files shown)


src/Command/CommandInterface.php
```diff
@@ -138,8 +138,8 @@ public function addPrimaryKey(string $table, string $name, array|string $columns
     /**
      * Creates an SQL command for changing the definition of a column.
      *
-     * @param string $table The table whose column is to be changed.
-     * @param string $column The name of the column to be changed.
+     * @param string $table The table whose column is to change.
+     * @param string $column The name of the column to change.
```
cebe, line 142: as far as I see, original is correct.  https://www.deepl.com/translator#de/en/Die%20Tabelle%2C%20dessen%20Spalte%20zu%20%C3%A4ndern%20ist.

src/QueryBuilder/DDLQueryBuilderInterface.php
```diff
@@ -274,8 +271,8 @@ public function dropCheck(string $table, string $name): string;
     /**
      * Builds an SQL statement for dropping a DB column.
      *
-     * @param string $table The table whose column is to dropp.
-     * @param string $column The name of the column to dropp.
+     * @param string $table The table whose column is to drop.
+     * @param string $column The name of the column to drop.
```
cebe, line 275:
```suggestion
     * @param string $column The name of the column to be dropped.
```

darkdef (approved):

src/QueryBuilder/DDLQueryBuilderInterface.php
samdark: Why? We're following MS style guide and it says to avoid passive voice.

src/Command/CommandInterface.php
samdark: It is as correct as new one but new one isn't passive voice.

PR #677, opened 2023-04-07
Remove unnecessary methods.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -624,18 +624,6 @@ class Query implements QueryInterface
         return $this;
     }
 
-    public function setJoin(array $value): static
-    {
-        $this->join = $value;
-        return $this;
-    }
-
-    public function setUnion(array $value): static
-    {
-        $this->union = $value;
-        return $this;
-    }
-
     public function shouldEmulateExecution(): bool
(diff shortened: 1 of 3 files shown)


samdark (approved):

darkdef (commented):
How about case 

```php
 $query = (new Query())->
...
many rows for build query
...
->join('t1')->join('t2');

$result = $query->all();
$cnt = $query->setJoin([])->count();
```

I'm not agree with this changes

cebe:
@darkdef that could be a use case, but in case you have Params added via join, the `setJoin` call will not remove them and the query will not work because more params are bound than are used in the query.

darkdef:
> @darkdef that could be a use case, but in case you have Params added via join, the `setJoin` call will not remove them and the query will not work because more params are bound than are used in the query.

More syntetical examples:
$query->join('LEFT', 't1', 't1.id=t2.id')
join can be without parameters

samdark:
OK. So we aren't removing methods.

PR #678, opened 2023-04-07
Improves the definition of the column builder.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


darkdef:
basic cases:
1. create table with not integer and not autoincrement pk
```
create table `name` (
str varchar(32) PRIMARY KEY
)
```

2. create table with another type of create PK
```
create table `name` (
`str` varchar(32),
PRIMARY KEY(`str`)
)
```
Current implementation not supports case 2. 
It's for think about it

cebe:
> Current implementation not supports case 2.

case 2 is mainly relevant for composite PK.

```sql
create table `name` (
`str` varchar(32) not null,
`owner_id` int not null
PRIMARY KEY(`str`, `owner_id`)
)
```

PR #682, opened 2023-04-09
Fix #635: Simplify cache implementation

### LINKED ISSUE #635: SchemaCache implementation is incomplete
### What steps will reproduce the problem?

`SchemaCache::getOrSet()` is only ever used as getter. In one place it is used as getOrSet() with `$value=null`, which will set a null value in case no cache entry exists.
It does not really make sense to have it implemented this way.

Also it claims to evaluate a callable for setting the value but that part is not implemented:

https://github.com/yiisoft/db/blob/3931c4937205afd4460dfe870f56e6cac72d1f7e/src/Cache/SchemaCache.php#L62-L97

In general the implementation of SchemaCache feels weird because it does not actually implement the caching. It is an object that holds the cache configuration and it wraps PSR Cache methods. But the caching behavior, i.e. handling the read and write of cache data is implemented in the Connection. Connection uses SchemaCache class to determine the configuration values and acts based on these. Instead most of the caching logic should be moved into SchemaCache.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #635

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,8 @@ composer.phar
 
 # phpunit itself is not needed
 phpunit.phar
-phpunit.result.cache
+.phpunit.result.cache
+tests/Support/runtime/
 
 # local phpunit config
 /phpunit.xml
--- a/src/Cache/SchemaCache.php
+++ b/src/Cache/SchemaCache.php
@@ -59,50 +59,37 @@ final class SchemaCache
     }
 
     /**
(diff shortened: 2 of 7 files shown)


cebe:
schemaCache->setEnable($value)

samdark:
Getter is already `isEnabled()`.

PR #703, opened 2023-04-13
foreach loop should use list()

KeremArdicli:
Also added extra information about the error to the catch block

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any

### DIFF
--- a/docs/en/queries/transactions.md
+++ b/docs/en/queries/transactions.md
@@ -45,7 +45,7 @@ try {
 } catch (Exception $e) {
     $transaction->rollBack();
     // to get a slightest info about the Exception
-    var_dump($e->errorInfo[2]);
+    var_dump($e->getMessage());
 }
 ```
 
--- a/docs/en/queries/transactions.md
+++ b/docs/en/queries/transactions.md
@@ -38,12 +38,14 @@ try {
         [3, 'db'],
     ];
         
-    foreach ($tags as [$id, $name] => $tag) {
+    foreach ($tags as list($id, $name)) {
         $insertTagCommand->execute();
         $insertPostTagCommand->execute();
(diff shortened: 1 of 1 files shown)


docs/en/queries/transactions.md
```diff
@@ -38,12 +38,14 @@ try {
         [3, 'db'],
     ];
         
-    foreach ($tags as [$id, $name] => $tag) {
+    foreach ($tags as list($id, $name)) {
```
samdark, line 41: Why? It's the same, isn't it?

docs/en/queries/transactions.md
```diff
@@ -38,12 +38,14 @@ try {
...
         $insertTagCommand->execute();
         $insertPostTagCommand->execute();
     }    
-} catch () {
+} catch (Exception $e) {
     $transaction->rollBack();
+    // to get a slightest info about the Exception
+    var_dump($e->getMessage());
```
samdark, line 48: That's a no go. Echoing exception isn't a good idea.

docs/en/queries/transactions.md
KeremArdicli: On my local env with win10, apache, PHP 8.1.1 it throws error. list() does not.

docs/en/queries/transactions.md
KeremArdicli: Yeah maybe commenting it and saying to use it only for development would be needed. It is a no go for prod.

samdark:
Thanks.

Issue #718, opened 2023-05-19
Use Param object instead of array


Tigrov:
https://github.com/yiisoft/db/blob/master/src/Schema/AbstractColumnSchema.php#L274
```php
if (
    is_array($value)
    && count($value) === 2
    && isset($value[1])
    && in_array($value[1], $this->getDataTypes(), true)
) {
    return new Param((string) $value[0], $value[1]);
}
```

Remove this code and use directly `new Param($value, $dataType)`

_To avoid problems with support values of array fields, eg:_
```php
$value = [$val1, $val2]; // where $val2 in array $this->getDataTypes()
// or
$value = ['key' => $val1, 1 => $val2];
```

The problem when you use array containing the value as the first element and the PDO type as the second you can get unexpected result for array fields, e.g.
```php
$db->createCommand()->createTable('{{%table}}', ['array' => 'json'])->execute(); // or array type instead of json ('integer []')
$db->createCommand()->insert('{{%table}}', ['array' => [1, 2]])->execute();
$result = $db->createCommand('SELECT * FROM {{%table}}')->queryScalar();
// expected $result === '[1, 2]'
```

Tigrov:
I found the array syntax to declare PDO parameter type was marked as deprecated.

https://github.com/SilverFire/yii2/blob/master/framework/UPGRADE.md?plain=1#L143
* `yii\db\PdoValue` class has been introduced to replace a special syntax that was used to declare PDO parameter type 
  when binding parameters to an SQL command, for example: `['value', \PDO::PARAM_STR]`.
  You should use `new PdoValue('value', \PDO::PARAM_STR)` instead. Old syntax will be removed in Yii 2.1.

PR #720, opened 2023-06-03
Column schema fix

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #718

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.0.1 under development
 
-- no changes in this release.
+- Bug #718: Сhanging the type casting check condition for value as array (@darkdef)
 
 ## 1.0.0 April 12, 2023
 
--- a/src/Schema/AbstractColumnSchema.php
+++ b/src/Schema/AbstractColumnSchema.php
@@ -203,7 +203,7 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
 
     public function phpTypecast(mixed $value): mixed
     {
-        return $this->typecast($value);
+        return $this->typecast($value, false);
     }
 
     public function precision(int|null $value): void
@@ -245,7 +245,7 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
      *
      * @return mixed The converted value.
      */
-    protected function typecast(mixed $value): mixed
+    protected function typecast(mixed $value, bool $dbCast = true): mixed
     {
         if (
             $value === ''
@@ -272,7 +272,8 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
         }
 
         if (
-            is_array($value)
+            $dbCast
+            && is_array($value)
             && count($value) === 2
             && isset($value[0], $value[1])
             && in_array($value[1], $this->getDataTypes(), true)
--- a/src/Schema/AbstractColumnSchema.php
+++ b/src/Schema/AbstractColumnSchema.php
@@ -271,14 +271,14 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
             return $value;
         }
 
-        if (
-            is_array($value)
-            && count($value) === 2
-            && isset($value[1])
-            && in_array($value[1], $this->getDataTypes(), true)
-        ) {
-            return new Param((string) $value[0], $value[1]);
-        }
+//        if (
+//            is_array($value)
+//            && count($value) === 2
+//            && isset($value[0], $value[1])
+//            && in_array($value[1], $this->getDataTypes(), true)
+//        ) {
+//            return new Param((string) $value[0], $value[1]);
+//        }
 
         switch ($this->phpType) {
             case SchemaInterface::PHP_TYPE_RESOURCE:
--- a/src/Schema/AbstractColumnSchema.php
+++ b/src/Schema/AbstractColumnSchema.php
@@ -271,14 +271,14 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
             return $value;
         }
 
-//        if (
-//            is_array($value)
-//            && count($value) === 2
-//            && isset($value[0], $value[1])
-//            && in_array($value[1], $this->getDataTypes(), true)
-//        ) {
-//            return new Param((string) $value[0], $value[1]);
-//        }
+        if (
+            is_array($value)
+            && count($value) === 2
+            && isset($value[0], $value[1])
+            && in_array($value[1], $this->getDataTypes(), true)
+        ) {
+            return new Param((string) $value[0], $value[1]);
+        }
 
         switch ($this->phpType) {
             case SchemaInterface::PHP_TYPE_RESOURCE:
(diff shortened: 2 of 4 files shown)


src/Schema/ColumnSchemaInterface.php
```diff
@@ -86,6 +86,8 @@ public function dbType(string|null $value): void;
      * directly in your code.
      *
      * If the value is `null` or an {@see Expression}, there will be no conversion.
+     * In exceptional case (with direct use) you can use array containing the value as the first element and the PDO type as the second
+     * element
```
samdark, line 90:
```suggestion
     * element.
```

samdark (approved):
Please add a line for CHANGELOG.

src/Schema/ColumnSchemaInterface.php
```diff
@@ -86,6 +86,8 @@ public function dbType(string|null $value): void;
      * directly in your code.
      *
      * If the value is `null` or an {@see Expression}, there will be no conversion.
+     * In exceptional case (with direct use) you can use array containing the value as the first element and the PDO type as the second
```
vjik, line 89: After this fix for array `['key' => 0, 1 => DataType::STRING]` returns array, but for array `[0, 2]`  returns `Param` object. It's unexpectedly. May be write in phpdoc "do not use directly"?

Protected method `typecast` uses in `phpTypecast()` method. Should we add notice and tests there also?

darkdef: Ok. Later

darkdef: Comment added.

Tigrov:
The problem with "(with direct use)" when you use array containing the value as the first element and the PDO type as the second you can get unexpected result for array fields, e.g.
```php
$db->createCommand()->createTable('{{%table}}', ['array' => 'json'])->execute(); // or array type instead of json ('integer []')
$db->createCommand()->insert('{{%table}}', ['array' => [1, 2]])->execute();
```
Maybe use a Param object where it is necessary instead of arrays?

Tigrov:
I think there is no real cases where you get array from db with PDO type as the second element. The idea to specify PDO type is to put a php value with correct PDO type into db but not vice versa.
It is not required to create a new Param object when you get values from db.

darkdef:
> It is not required to create a new Param object when you get values from db.
Yes and i was add fix for it behaviour. It's for futuring - non PDO implementations.

Now corrects for your issue?

Tigrov:
> Now corrects for your issue?

```php
$db->createCommand()->createTable('{{%table}}', ['array' => 'json'])->execute(); // or array type instead of json ('integer []')
$db->createCommand()->insert('{{%table}}', ['array' => [1, 2]])->execute();
$result = $db->createCommand('SELECT * FROM {{%table}}')->queryScalar();
// expected $result === '[1, 2]'
```

Will this example work correct?

darkdef:
> 



> Will this example work correct?

In postgresql - yes

```json
string(6) "[1, 2]"
```
But and before fix all been ok (in pgsql)

Tigrov:
Yes, you are right. Thanks.

What about the comment https://github.com/yiisoft/db/issues/718#issuecomment-1576043917

darkdef:
> What about the comment [#718 (comment)](https://github.com/yiisoft/db/issues/718#issuecomment-1576043917)

In other PR. @terabytesoftw work for remove legacy behavior.

src/Schema/AbstractColumnSchema.php
```diff
@@ -245,7 +245,7 @@ public function unsigned(bool $value): void
      *
      * @return mixed The converted value.
      */
-    protected function typecast(mixed $value): mixed
+    protected function typecast(mixed $value, bool $dbCast = true): mixed
```
vjik, line 248: Need fix phpdoc of this method.

src/Schema/AbstractColumnSchema.php
```diff
@@ -242,10 +242,11 @@ public function unsigned(bool $value): void
      * If the value is null or an {@see Expression}, it won't be converted.
      *
      * @param mixed $value The value to be converted.
+     * @param bool $dbCast Flag for detect direction of type casting (read or write to db)
```
samdark, line 245:
```suggestion
     * @param bool $dbCast Whether to detect direction of type casting (read or write to db).
```

CHANGELOG.md
```diff
@@ -2,8 +2,8 @@
 
 ## 1.0.1 under development
 
-- no changes in this release.
+- Bug #718: Сhanging the type casting check condition for value as array (@darkdef)
```
samdark, line 5:
```suggestion
- Bug #718: Fix passing value => PDO type pair to `AbstractColumnSchema::typecast()` (@darkdef)
```

src/Schema/ColumnSchemaInterface.php
```diff
@@ -86,6 +86,8 @@ public function dbType(string|null $value): void;
      * directly in your code.
      *
      * If the value is `null` or an {@see Expression}, there will be no conversion.
+     * In exceptional case (with direct use) you can use array containing the value as the first element and the PDO
+     * type as the second element.
```
samdark, line 90:
```suggestion
     * When using the method directly you can pass an array containing the value
     * as the first element and the PDO type as the second element.
```

src/Schema/ColumnSchemaInterface.php
```diff
@@ -265,6 +267,8 @@ public function phpType(string|null $value): void;
      * Converts the input value according to {@see phpType} after retrieval from the database.
      *
      * If the value is `null` or an {@see Expression}, there is no conversion.
+     * In exceptional case (with direct use) you can use array containing the value as the first element and the PDO
+     * type as the second element.
```
samdark, line 271:
```suggestion
     * When using the method directly you can pass an array containing the value
     * as the first element and the PDO type as the second element.
```

samdark (approved):

Tigrov:
Maybe to move the code 

```php
if (
    $dbCast
    && is_array($value)
    && count($value) === 2
    && isset($value[0], $value[1])
    && in_array($value[1], $this->getDataTypes(), true)
) {
    return new Param((string) $value[0], $value[1]);
}
```

to 

```php
public function dbTypecast(mixed $value): mixed
{
...
}
```

instead of adding a new attribute `bool $dbCast = true`?

terabytesoftw:
> Maybe to move the code
> 
> ```
> if (
>     $dbCast
>     && is_array($value)
>     && count($value) === 2
>     && isset($value[0], $value[1])
>     && in_array($value[1], $this->getDataTypes(), true)
> ) {
>     return new Param((string) $value[0], $value[1]);
> }
> ```
> 
> to
> 
> ```
> public function dbTypecast(mixed $value): mixed
> {
> ...
> }
> ```
> 
> instead of adding a new attribute `bool $dbCast = true`?

In my opinion the best solution is to remove this strange behavior, for me it does not make sense to cast only two values of an array, since in all other scenarios it will fail, it is best to return the values as each dbms returns them, this does not affect at all the operation of all db packages, which i have tested without this code and they all work perfectly.

darkdef:
Not actually - https://github.com/yiisoft/db/pull/722

PR #723, opened 2023-06-27
Fix bit type

Tigrov:
For `yiisoft/db-pgsql`

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Schema/AbstractSchema.php
+++ b/src/Schema/AbstractSchema.php
@@ -409,6 +409,7 @@ abstract class AbstractSchema implements SchemaInterface
             SchemaInterface::TYPE_SMALLINT => SchemaInterface::PHP_TYPE_INTEGER,
             SchemaInterface::TYPE_INTEGER => SchemaInterface::PHP_TYPE_INTEGER,
             SchemaInterface::TYPE_BIGINT => SchemaInterface::PHP_TYPE_INTEGER,
+            SchemaInterface::TYPE_BIT => SchemaInterface::PHP_TYPE_INTEGER,
             SchemaInterface::TYPE_BOOLEAN => SchemaInterface::PHP_TYPE_BOOLEAN,
             SchemaInterface::TYPE_DECIMAL => SchemaInterface::PHP_TYPE_DOUBLE,
             SchemaInterface::TYPE_FLOAT => SchemaInterface::PHP_TYPE_DOUBLE,
--- a/src/Schema/SchemaInterface.php
+++ b/src/Schema/SchemaInterface.php
@@ -207,6 +207,10 @@ interface SchemaInterface extends ConstraintSchemaInterface
      */
     public const TYPE_BINARY = 'binary';
     /**
+     * Define the abstract column type as `bit`.
+     */
+    public const TYPE_BIT = 'bit';
+    /**
      * Define the abstract column type as `boolean`.
      */
     public const TYPE_BOOLEAN = 'boolean';


samdark (approved):

src/Schema/AbstractSchema.php
```diff
@@ -409,6 +409,7 @@ protected function getColumnPhpType(ColumnSchemaInterface $column): string
             SchemaInterface::TYPE_SMALLINT => SchemaInterface::PHP_TYPE_INTEGER,
             SchemaInterface::TYPE_INTEGER => SchemaInterface::PHP_TYPE_INTEGER,
             SchemaInterface::TYPE_BIGINT => SchemaInterface::PHP_TYPE_INTEGER,
+            SchemaInterface::TYPE_BIT => SchemaInterface::PHP_TYPE_INTEGER,
```
vjik, line 412: In MSSQL BIT is 1, 0, or NULL only. Can it be problem?

vjik: May be use name "BIT_STRING"?

Tigrov: The changes are only for `yiisoft/db-pgsql`

For `yiisoft/db-mssql` it is as before:

https://github.com/yiisoft/db-mssql/blob/master/src/Schema.php#L74C9-L74C38
```php
'bit' => self::TYPE_SMALLINT,
```

I do not know any issues with type `bit` in MSSQL but if it is required it is possible to change it to `'bit' => self::TYPE_BIT,` and cast `bit` value in `ColumnSchema::dbTypecast()` or `ColumnSchema::phpTypecast()`.

terabytesoftw: As it is not a global `type` that behaves the same for all `dbms`, i think it would be a good idea to map it in `db-pgsql`, and each `dbms` will have its own behavior, as it should be.

src/Schema/SchemaInterface.php
```diff
@@ -206,6 +206,10 @@ interface SchemaInterface extends ConstraintSchemaInterface
      * Define the abstract column type as `binary`.
      */
     public const TYPE_BINARY = 'binary';
+    /**
```
terabytesoftw, line 209: As it is not a global `type` that behaves the same for all `dbms`, i think it would be a good idea to map it in `db-pgsql`, and each `dbms` will have its own behavior, as it should be.

src/Schema/SchemaInterface.php
```diff
@@ -206,6 +206,10 @@ interface SchemaInterface extends ConstraintSchemaInterface
...
+     * Define the abstract column type as `bit`.
+     */
+    public const TYPE_BIT = 'bit';
```
terabytesoftw, line 212: As it is not a global type that behaves the same for all dbms, i think it would be a good idea to map it in db-pgsql, and each dbms will have its own behavior, as it should be.

Tigrov: `public const TYPE_JSONB = 'json';` is also for `db-pgsql` only.
Should we also move it into `Yiisoft\Db\Pgsql\Schema`?

terabytesoftw: I think so, basically we should have mapped in db, only the generic types, and the types with specific behaviors in each implementation.

Tigrov:
Code moved to `yiisoft/db-pgsql`

terabytesoftw:
Thank you. 👍

PR #724, opened 2023-07-04
Typecast refactoring

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Schema/AbstractColumnSchema.php
+++ b/src/Schema/AbstractColumnSchema.php
@@ -253,14 +253,12 @@ abstract class AbstractColumnSchema implements ColumnSchemaInterface
             return null;
         }
 
-        if (
-            $value instanceof ExpressionInterface
-            || gettype($value) === $this->phpType
-        ) {
+        if ($value instanceof ExpressionInterface) {
             return $value;
         }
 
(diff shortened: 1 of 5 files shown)


terabytesoftw (approved):

src/Schema/AbstractColumnSchema.php
```diff
@@ -241,60 +241,39 @@ public function unsigned(bool $value): void
...
-                 *
-                 * @link https://github.com/yiisoft/yii2/issues/9006
-                 */
-                return (bool) $value && $value !== "\0";
-            case SchemaInterface::PHP_TYPE_DOUBLE:
-                return (float) $value;
-        }
-
-        return $value;
+        return match (true) {
```
darkdef, line 244: If you want to reduce cognitive complexity - don't use match(true).
Do refactoring with remove switch/case and simplify conditions in "if".
Early exit work fastest.

Tigrov: Done

darkdef:
Thanks

PR #744, opened 2023-08-17
Different tables in columns

Tigrov:
What is the correct behavior?
1. Filter columns from other tables
2. Let them be in the query

The test example is about the code:
```php
$table = '{{%type}}';
$columns = ['{{%type}}.[[bool_col]]', '{{%another_table}}.[[bool_col2]]'];
$values = [[true, false]];

$command->batchInsert($table, $columns, $values);

// INSERT INTO {{%type}} ([[bool_col]], [[bool_col2]]) VALUES (:qp0, :qp1)
```

The same for methods `batchInsert()`, `insert()`, `update()`, `upsert()`

### DIFF
--- a/tests/Provider/QueryBuilderProvider.php
+++ b/tests/Provider/QueryBuilderProvider.php
@@ -213,6 +213,18 @@ class QueryBuilderProvider
                 ),
                 [':qp0' => null],
             ],
+            'different tables in columns' => [
+                '{{%type}}',
+                ['{{%type}}.[[bool_col]]', '{{%another_table}}.[[bool_col2]]'],
+                [[true, false]],
+                'expected' => DbHelper::replaceQuotes(
+                    <<<SQL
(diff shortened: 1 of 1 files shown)


terabytesoftw:
If you are looking at the test now, of course it is a bug, since it is specifying another table.

Tigrov:
@terabytesoftw
> If you are looking at the test now, of course it is a bug.

Or we can mention about this in the doc for these methods:
`@param string[] $columns The column names of the table.`

Now it is
`@param string[] $columns The column names.`

terabytesoftw:
> @terabytesoftw
> 
> > If you are looking at the test now, of course it is a bug.
> 
> Or we can mention about this in the doc for these methods: `@param string[] $columns The column names of the table.`
> 
> Now it is `@param string[] $columns The column names.`

If that’s all right.

Tigrov:
Solved by #746

Issue #737, opened 2023-08-07
Suggestion to speed up the casting of values


Tigrov:
Suggestion to predefine cast methods in a separate class (e.g. `Typecast`)
https://github.com/Tigrov/db-pgsql/blob/speedup_typecast_predefined/src/Typecast.php

Then initialize the predefined methods in `ColumnSchema` objects
https://github.com/Tigrov/db-pgsql/blob/speedup_typecast_predefined/src/Schema.php#L819

And use the predefined methods when casting values
https://github.com/Tigrov/db-pgsql/blob/speedup_typecast_predefined/src/ColumnSchema.php#L84
https://github.com/Tigrov/db-pgsql/blob/speedup_typecast_predefined/src/ColumnSchema.php#L131

This increase the speed of casting values by an average of **30-40%**

More information and benchmarks
https://github.com/Tigrov/db-pgsql/tree/speedup_typecast_predefined

### Update:

If `phpTypecast()` casts values retrieved from DB only, some checks can be removed, this will increase the speed by **10-20%** more.

| Q                | A
| ---------------- | ---
| Version          | 1.1.0
| PHP version      | any
| Operating system | any

vjik:
I like idea clearly separate type casting "db → php" and "php → db".

May be split typecast class to two classes: 
- `DbTypeCaster` - converts a value from its PHP representation to a database-specific representation;
- `PhpTypeCaster` - converts the input value to PHP type after retrieval from the database.

These classes should be placed in DB implementations packages and uses into `ColumnSchema`. May be in Yii DB create `AbstractTypeCaster`.

-----

Also has idea create classes for each column types. In this case type casting should be implemented into this clasess.

Tigrov:
> Also has idea create classes for each column types. In this case type casting should be implemented into this clasess.

Also thought about this, looks better 👍

> These classes should be placed in DB implementations packages and uses into `ColumnSchema`. May be in Yii DB create `AbstractTypeCaster`

👍

Sure, this just an example.

Tigrov:
Realized example of creating classes for each column type with optimization of `phpTypecast()` for values retrieved from DB only

## Results for `dbTypecast()`

Predefined methods are **50%** faster
Column type classes are **165%** faster

## Results for `phpTypecast()`

Predefined methods are **60%** faster
Column type classes are **200%** faster

More info https://github.com/Tigrov/db-pgsql/tree/speedup_typecast_columns

PR #752, opened 2023-09-01
ColumnSchema classes for performance of typecasting

Tigrov:
### Related PRs ###
* yiisoft/db-mysql#303
* yiisoft/db-pgsql#315
* yiisoft/db-sqlite#273
* yiisoft/db-mssql#277
* yiisoft/db-oracle#236

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #737, yiisoft/db-mysql#292


src/Schema/Column/AbstractColumnSchema.php
```diff
@@ -0,0 +1,219 @@
...
+    private string|null $extra = null;
+    private bool $isPrimaryKey = false;
+    private string|null $phpType = null;
+    private int|null $precision = null;
+    private int|null $scale = null;
+    private int|null $size = null;
+    private string $type = '';
+    private bool $unsigned = false;
+
+    public function __construct(private string $name)
```
darkdef, line 51: How about add $type and $phptype params to constructor?

And use in child class:
parent::__costruct($name, SchemaInterface::TYPE_BIGINT, SchemaInterface::PHP_TYPE_BIGINT) etc

darkdef: Setters will be removed

src/Schema/AbstractSchema.php
```diff
@@ -422,6 +459,18 @@ protected function getColumnPhpType(ColumnSchemaInterface $column): string
         };
     }
 
+    protected function createPhpTypeColumnSchema(string $phpType, string $name): ColumnSchemaInterface
```
darkdef, line 462: It's bad name for function

src/Schema/AbstractSchema.php
```diff
@@ -394,24 +402,53 @@ protected function findTableNames(string $schema): array
...
+        $phpType = $this->getColumnPhpType($type);
+        $isUnsigned = !empty($info['unsigned']);
+
+        if (
+            $isUnsigned && PHP_INT_SIZE === 4 && $type === SchemaInterface::TYPE_INTEGER
+            || ($isUnsigned || PHP_INT_SIZE !== 8) && $type === SchemaInterface::TYPE_BIGINT
+        ) {
+            $column = new BigIntColumnSchema($name);
+        } else {
+            $column = $this->createPhpTypeColumnSchema($phpType, $name);
```
darkdef, line 427: Very strange behavior. Why BigIntColumn not in this function ?

Tigrov: Creating `BigIntColumnSchema` depends on `$isUnsigned`, `PHP_INT_SIZE`, `$type`
`createPhpTypeColumnSchema()` depends only on `$phpType`

src/Schema/AbstractSchema.php
Tigrov: Any suggestions for a name?

src/Schema/Column/AbstractColumnSchema.php
Tigrov: Done

src/Schema/AbstractSchema.php
Tigrov: Renamed to `createColumnSchemaFromPhpType()`

src/Schema/AbstractSchema.php
Tigrov: Done

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Enh #806: Build `Expression` instances inside `Expression::$params` when build a query using `QueryBuilder` (@Tigrov)
 - Enh #766: Allow `ColumnInterface` as column type. (@Tigrov)
 - Bug #828: Fix `float` type when use `AbstractCommand::getRawSql()` method (@Tigrov)
+- Enh #752: Implement `ColumnSchemaInterface` classes according to the data type of database table columns
```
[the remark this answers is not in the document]
Tigrov: Done

src/Schema/Column/BigIntColumnSchema.php
```diff
@@ -0,0 +1,49 @@
...
+    public function dbTypecast(mixed $value): int|string|ExpressionInterface|null
+    {
+        if (is_int($value)) {
+            return $value;
+        }
+
+        return match ($value) {
+            null, '' => null,
+            false => 0,
+            default => $value instanceof ExpressionInterface
```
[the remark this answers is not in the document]
Tigrov: This will not be optimal. One of the optimization ideas is to move this check to the last plan. First we check standard situations, then the rest. In the case with `bigint` we check if the value is `string`, `null` or `int` and if not we check other cases. Refactored.

src/Schema/Column/BigIntColumnSchema.php
```diff
@@ -0,0 +1,49 @@
...
+            'string' => $value === '' ? null : (
+                $value <= PHP_INT_MAX && $value >= PHP_INT_MIN
+                    ? (int) $value
+                    : $value
+            ),
+            'NULL' => null,
+            'integer' => $value,
+            'boolean' => $value ? 1 : 0,
+            default => $value instanceof ExpressionInterface ? $value : (
+                ($val = (string) $value) <= PHP_INT_MAX && $val >= PHP_INT_MIN
```
vjik, line 34: It's duplicate code with string case, may be extract it to separate method?

src/Schema/Column/BigIntColumnSchema.php
```diff
@@ -0,0 +1,49 @@
...
+                    ? (int) $val
+                    : $val
+            ),
+        };
+    }
+
+    public function phpTypecast(mixed $value): string|null
```
vjik, line 41:
```suggestion
    public function phpTypecast(mixed $value): string|int|null
```

Can we make it?

Tigrov: I thought about doing this before, but then I changed my mind.

For PHP there is no difference between `string` and `int` numbers, but supporting both types may be cause of cases when it is necessary to do more checks on whether the value is `string` or `int` or every time cast value to `string`.

The value from the database is assumed to be a large number and needs to be worked with as a large number, even if it is within the limits of the `int` type.

src/Schema/Column/BigIntColumnSchema.php
Tigrov: The difference is to cast the value to string before check. And calling an additional method will slow down performance. Not so critical, but also not particularly necessary.

samdark (approved):

PR #766, opened 2023-10-26
Allow `ColumnInterface` as column type

Tigrov:
First step to solve yiisoft/db-migration#218, yiisoft/db-migration#219 and other

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -


src/QueryBuilder/AbstractDDLQueryBuilder.php
```diff
@@ -169,14 +169,15 @@ public function createTable(string $table, array $columns, string $options = nul
     {
         $cols = [];
 
-        /** @psalm-var string[] $columns */
+        /** @psalm-var array<string, ColumnInterface>|string[] $columns */
```
vjik, line 172:
```suggestion
        /** @psalm-var array<array-key, ColumnInterface|string> $columns */
```

src/QueryBuilder/AbstractDDLQueryBuilder.php
```diff
@@ -169,14 +169,15 @@ public function createTable(string $table, array $columns, string $options = nul
...
         foreach ($columns as $name => $type) {
             if (is_string($name)) {
                 $cols[] = "\t"
                     . $this->quoter->quoteColumnName($name)
                     . ' '
                     . $this->queryBuilder->getColumnType($type);
             } else {
+                /** @psalm-var string $type */
                 $cols[] = "\t" . $type;
```
vjik, line 181:
```suggestion
                $cols[] = "\t" . $type;
```

if `$name` is integer, then here will throw exception, becuase `ColumnInterface` is not stringable.

tests/Db/Command/CommandTest.php
```diff
@@ -39,17 +40,20 @@ public function testAddCheck(): void
...
+    /** @dataProvider \Yiisoft\Db\Tests\Provider\CommandProvider::columnTypes */
+    public function testAddColumn(ColumnInterface|string $type): void
     {
         $db = $this->getConnection();
 
         $command = $db->createCommand();
-        $sql = $command->addColumn('table', 'column', SchemaInterface::TYPE_INTEGER)->getSql();
+        $sql = $command->addColumn('table', 'column', $type)->getSql();
+
+        $type = $db->getQueryBuilder()->getColumnType($type);
```
vjik, line 51: Suggest use another name for variable. $type here and $type in method parameters are confused =)

vjik (changes requested):

src/QueryBuilder/AbstractDDLQueryBuilder.php
Tigrov: If column is `ColumnInterface` it should have name as an array key => `array<string, ColumnInterface>`
If it is `string` then the array key can be `int|string` => `string[]`

src/QueryBuilder/AbstractDDLQueryBuilder.php
Tigrov: `ColumnInterface` can be only with string `$name`

src/QueryBuilder/AbstractDDLQueryBuilder.php
vjik: Can you add columns format to phpdoc in `DDLQueryBuilderInterface`?

vjik (approved):

darkdef (approved):

rustamwin (approved):

vjik (changes requested):
Need line in changelog only

vjik (approved):

PR #775, opened 2023-11-12
Fix BC issues

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/Query/QueryPartsInterface.php
+++ b/src/Query/QueryPartsInterface.php
@@ -645,11 +645,10 @@ interface QueryPartsInterface
      * Prepends an SQL statement using `WITH` syntax.
      *
      * @param QueryInterface|string $query The SQL statement to append using `UNION`.
-     * @param ExpressionInterface|string $alias The query alias in `WITH` construction.
-     * To specify the alias in plain SQL, you may pass an instance of {@see ExpressionInterface}.
+     * @param string $alias The query alias in `WITH` construction.
      * @param bool $recursive Its `true` if using `WITH RECURSIVE` and `false` if using `WITH`.
      */
-    public function withQuery(QueryInterface|string $query, ExpressionInterface|string $alias, bool $recursive = false): static;
+    public function withQuery(QueryInterface|string $query, string $alias, bool $recursive = false): static;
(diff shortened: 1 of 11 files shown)


src/Query/QueryPartsInterface.php
```diff
@@ -645,11 +645,10 @@ public function where(array|string|ExpressionInterface|null $condition, array $p
      * Prepends an SQL statement using `WITH` syntax.
      *
      * @param QueryInterface|string $query The SQL statement to append using `UNION`.
-     * @param ExpressionInterface|string $alias The query alias in `WITH` construction.
-     * To specify the alias in plain SQL, you may pass an instance of {@see ExpressionInterface}.
+     * @param string $alias The query alias in `WITH` construction.
      * @param bool $recursive Its `true` if using `WITH RECURSIVE` and `false` if using `WITH`.
      */
-    public function withQuery(QueryInterface|string $query, ExpressionInterface|string $alias, bool $recursive = false): static;
+    public function withQuery(QueryInterface|string $query, string $alias, bool $recursive = false): static;
```
vjik, line 651: Please, create issue for revert this change before next major version.

src/QueryBuilder/AbstractDMLQueryBuilder.php
```diff
@@ -49,7 +50,7 @@ public function __construct(
     ) {
     }
 
-    public function batchInsert(string $table, array $columns, iterable $rows, array &$params = []): string
+    public function batchInsert(string $table, array $columns, iterable|Generator $rows, array &$params = []): string
```
vjik, line 53: Removing `Generator` in this case is not break BC. See https://3v4l.org/SENSO

vjik (changes requested):

src/Query/QueryPartsInterface.php
vjik: https://github.com/yiisoft/db/issues/776

vjik (approved):

PR #778, opened 2023-11-17
Deprecate unnecessary argument `$rawSql` of `AbstractCommand::internalExecute()`

Tigrov:
The optimization also removes extra call `$this->getRawSql()`.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## 1.2.1 under development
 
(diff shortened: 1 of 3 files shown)


vjik (changes requested):
Add line to changelog only

PR #781, opened 2023-11-22
Do not call `CommandInterface::getRawSql()` if no `logger` or `profiler`

Tigrov:
Optimization

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/Driver/Pdo/AbstractPdoCommand.php
+++ b/src/Driver/Pdo/AbstractPdoCommand.php
@@ -257,8 +257,6 @@ abstract class AbstractPdoCommand extends AbstractCommand implements PdoCommandI
 
     /**
      * Logs the current database query if query logging is on and returns the profiling token if profiling is on.
-     *
-     * @deprecated Use {@see logger} instead. Will be removed in version 2.0.0.
      */
     protected function logQuery(string $rawSql, string $category): void
     {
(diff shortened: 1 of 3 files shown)


vjik (commented):
If add result caching to `getRawSql()` method, then we can reafactor code so:

```php
$this->profiler?->begin($this->getRawSql(), $queryContext);
```

If profile is null then `$this->getRawSql()` don't call.

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -265,14 +265,18 @@ protected function logQuery(string $rawSql, string $category): void
 
     protected function queryInternal(int $queryMode): mixed
     {
-        $rawSql = $this->getRawSql();
         $logCategory = self::class . '::' . $this->getQueryMode($queryMode);
 
-        $this->logQuery($rawSql, $logCategory);
+        $this->logger?->log(LogLevel::INFO, $rawSql = $this->getRawSql(), [$logCategory]);
```
vjik, line 270: If `logQuery()` not used in adapters, we can mark his deprecated and remove before next major release.

vjik (approved):

xepozz (approved):

vjik (approved):

PR #784, opened 2023-11-24
Specify result type of `ConstraintSchemaInterface::getTableIndexes()` method to `IndexConstraint[]`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/Schema/AbstractSchema.php
+++ b/src/Schema/AbstractSchema.php
@@ -10,6 +10,7 @@ use Yiisoft\Db\Cache\SchemaCache;
 use Yiisoft\Db\Command\DataType;
 use Yiisoft\Db\Connection\ConnectionInterface;
 use Yiisoft\Db\Constraint\Constraint;
+use Yiisoft\Db\Constraint\IndexConstraint;
 use Yiisoft\Db\Exception\NotSupportedException;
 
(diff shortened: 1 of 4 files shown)


src/Schema/AbstractSchema.php
```diff
@@ -261,7 +262,7 @@ public function getTableForeignKeys(string $name, bool $refresh = false): array
      */
     public function getTableIndexes(string $name, bool $refresh = false): array
     {
-        /** @psalm-var mixed $tableIndexes */
+        /** @var IndexConstraint[]|null $tableIndexes */
```
Tigrov, line 265: Are there cases when it returns `null`?

vjik: Hm... I don't know. If no, then we can remove check "is_array".

Tigrov: > Hm... I don't know. If no, then we can remove check "is_array".

Yes, looks like this.

vjik: Done

Tigrov (approved):

Issue #772, opened 2023-11-10
Allow use `DMLQueryBuilderInterface::batchInsert()` without `$columns`


Tigrov:
`$columns` can be obtained
1. from `$rows` if values are associative array
2. from `$table` if table shema found
3. or generate insert without columns if no columns found

Need to replace arguments 
from `DMLQueryBuilderInterface::batchInsert(string $table, array $columns, iterable $rows, array &$params = [])`
to `DMLQueryBuilderInterface::batchInsert(string $table, iterable $rows, array $columns = [], array &$params = [])`

This will allow to use the method this way `batchInsert($table, $rows)`

vjik:
1 - 👍
2 - but we don't know order of fields in rows, also rows may contain not all fields. Seems, it's can lead to errors in user code.
3 - 👍

Tigrov:
2. This is almost the same case as case 3, but we can get fields from table schema for casting type of values

vjik:
> If no colums passed then means columns have order as in table schema

But why need columns in this case? Isn't that the equivalent to way 3?

Tigrov:
> But why need columns in this case? Isn't that the equivalent to way 3?

Columns are needed for type casting and seems not all DBMS supports partial lists of values, but we can support them

PR #795, opened 2024-01-03
Batch insert with empty columns

Tigrov:
Prepare for #772

Allow to use `DMLQueryBuilderInterface::batchInsert($table, [], $rows)` method with empty columns

### Related PR
* yiisoft/db-oracle#251

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️/❌
| Breaks BC?    | ❌
| Fixed issues  | -


tests/Provider/QueryBuilderProvider.php
```diff
@@ -171,11 +171,11 @@ public static function batchInsert(): array
                 [['no columns passed']],
                 'expected' => DbHelper::replaceQuotes(
                     <<<SQL
-                    INSERT INTO [[customer]] () VALUES (:qp0)
+                    INSERT INTO [[customer]] ([[id]]) VALUES (:qp0)
```
darkdef, line 174: Why inserted in "id" ? I think - this incorrect behavior

tests/Provider/QueryBuilderProvider.php
```diff
@@ -171,11 +171,11 @@ public static function batchInsert(): array
...
                     SQL,
                     static::$driverName,
                 ),
-                'expectedParams' => [':qp0' => 'no columns passed'],
+                'expectedParams' => [':qp0' => 0],
```
darkdef, line 178: Why 'no columns passed' casted to int ?

Tigrov: The first column from the table `customer` is `id` and it is `int`

tests/Provider/QueryBuilderProvider.php
darkdef: I'm not agree with this behavior

darkdef (changes requested):

darkdef (approved):

vjik (changes requested):
Need line in changelog

vjik (approved):

PR #798, opened 2024-01-21
Restore Query::one interface

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️/❌


src/Query/QueryInterface.php
```diff
@@ -249,10 +249,10 @@ public function getWithQueries(): array;
      * @throws InvalidConfigException
      * @throws Throwable
      *
-     * @return array|null The first row (in terms of an array) of the query result. It returns `null` if the query
-     * results in nothing.
+     * @return array|object|null The first row (as array or object) of the query result.
+     * It returns `null` if the query results in nothing.
      */
-    public function one(): array|null;
+    public function one(): array|object|null;
```
vjik, line 255: Should we make same change in `all()` method and in `Query` class?

Tigrov (commented):
`BatchQueryResult::$value` and related methods also should have the type `array|object|null`
https://github.com/yiisoft/db/blob/ef81e8b7749525030dba95060b71a9f490fe2784/src/Query/BatchQueryResult.php#L43

src/Query/QueryInterface.php
darkdef: fixed

darkdef:
> `BatchQueryResult::$value` and related methods also should have the type `array|object|null`

It's maybe incorrect for query with one column as scalar type.

Tigrov:
> > `BatchQueryResult::$value` and related methods also should have the type `array|object|null`
> 
> It's maybe incorrect for query with one column as scalar type.

This should be an array ($row) or array of arrays ($rows) or null, but seems this is not related with the PR.

Tigrov (changes requested):
Needs line to changelog

src/Query/QueryInterface.php
```diff
@@ -50,7 +50,8 @@ public function addParams(array $params): static;
      * @throws InvalidConfigException
      * @throws Throwable
      *
-     * @return array[] The query results. If the query results in nothing, it returns an empty array.
+     * @return array The query results. If the query results in nothing, it returns an empty array.
+     * @psalm-return array[]|object[]
```
vjik, line 54:
```suggestion
     * @return array[]|object[] The query results. If the query results in nothing, it returns an empty array.
     * @psalm-return list<array|object>
```

src/Command/CommandInterface.php
```diff
@@ -631,10 +631,10 @@ public function queryColumn(): array;
      * @throws Exception
      * @throws Throwable If execution failed.
      *
-     * @return array|null The first row (in terms of an array) of the query result. Null if the query
+     * @return array|object|null The first row (in terms of an array) of the query result. Null if the query
```
vjik, line 634:
```suggestion
     * @return array|object|null The first row (as array or object) of the query result. Null if the query
```

vjik (changes requested):
Need also fix type of `CommandInterface::queryAll()`

vjik:
> Needs line to changelog

And fill `UPGRADE.md` too

samdark (approved):

Tigrov (approved):

Tigrov:
> Do we need also fix type of `CommandInterface::queryAll()`?

Currently it looks excessive and needs refactoring `DbArrayHelper` for supporting objects but it looks redundant.

vjik (approved):

PR #800, opened 2024-01-22
Base refresh materialized view

Gerych1984:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -833,4 +833,14 @@ interface CommandInterface
         bool|array $updateColumns = true,
         array $params = []
     ): static;
+
+    /**
+     *
+     * Execute `REFRESH MATERIALIZED VIEW` command
+     * @param string $viewName
+     * @param bool|null $concurrently If `null` then auto choice from depends on DBMS
+     * @param bool|null $withData
+     * @return bool
+     */
+    public function refreshMaterializedView(string $viewName, ?bool $concurrently = null, ?bool $withData = null): bool;
 }
--- a/src/QueryBuilder/DDLQueryBuilderInterface.php
+++ b/src/QueryBuilder/DDLQueryBuilderInterface.php
@@ -424,4 +424,15 @@ interface DDLQueryBuilderInterface
      * Note: The method will quote the `table` parameter before using it in the generated SQL.
      */
     public function truncateTable(string $table): string;
+
+    /**
+     * Refresh materialized view
+     *
+     * @param string $viewName The name of the view to refresh.
+     * @param bool $concurrently Refresh the materialized view without locking out concurrent selects on the materialized view.
+     * @param bool|null $withData When `true` then the backing query is executed to provide the new data. Otherwise if `false` then no new data is generated and the materialized view is left in an unscannable state
+     *
+     * @return string The `REFRESH MATERIALIZED VIEW` SQL statement
+     */
+    public function refreshMaterializedView(string $viewName, bool $concurrently = false, ?bool $withData = null): string;
 }
--- a/tests/Db/QueryBuilder/QueryBuilderTest.php
+++ b/tests/Db/QueryBuilder/QueryBuilderTest.php
@@ -178,7 +178,7 @@ final class QueryBuilderTest extends AbstractQueryBuilderTest
 
         $this->expectException(Exception::class);
 
-        $expression = new class () implements ExpressionInterface {
+        $expression = new class() implements ExpressionInterface {
         };
         $qb = $db->getQueryBuilder();
         $qb->getExpressionBuilder($expression);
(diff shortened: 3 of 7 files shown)


Tigrov (commented):
I think materialized views should be added in complex. With `create`, `alter`, `drop` and `refresh` methods. And take into account features of all supported DBMSs.

So, we should realize all these methods for all supported DBMSs which support materialized views.

Gerych1984:
> I think materialized views should be added in complex. With `create`, `alter`, `drop` and `refresh` methods. And take into account features of all supported DBMSs.
> 
> So, we should realize all these methods for all supported DBMSs which support materialized views.

That's what this PR is made for. Without it, i can't go any further

src/Command/CommandInterface.php
```diff
@@ -833,4 +833,14 @@ public function upsert(
...
+
+    /**
+     *
+     * Execute `REFRESH MATERIALIZED VIEW` command
+     * @param string $viewName
+     * @param bool|null $concurrently If `null` then auto choice from depends on DBMS
+     * @param bool|null $withData
+     * @return bool
+     */
+    public function refreshMaterializedView(string $viewName, ?bool $concurrently = null, ?bool $withData = null): bool;
```
Tigrov, line 845:
```suggestion
    public function refreshMaterializedView(string $viewName, bool $concurrently = false, ?bool $withData = null): bool;
```

Why the interface is different from `DDLQueryBuilderInterface::refreshMaterializedView()`?
Looks like should be the same.

When execute `$command->refreshMaterializedView($viewName);` expect this is not concurently.

tests/Db/QueryBuilder/QueryBuilderTest.php
```diff
@@ -310,4 +310,64 @@ public function testUpsertExecute(
...
+     * @param bool $concurrently
+     * @param bool|null $withData
+     * @return void
+     */
+    public function testRefreshMaterializedView(string $viewName, bool $concurrently, ?bool $withData, string $expected): void
+    {
+        $db = $this->getConnection();
+        $driver = $db->getDriverName();
+        $sql = $db->getQueryBuilder()->refreshMaterializedView($viewName, $concurrently, $withData);
+        $actual = DbHelper::replaceQuotes($sql, $driver);
```
Tigrov, line 368:
```suggestion
        $actual = $db->getQueryBuilder()->refreshMaterializedView($viewName, $concurrently, $withData);
```

Result from `$db->getQueryBuilder()->refreshMaterializedView()` already quoted acording to the driver

tests/Db/QueryBuilder/QueryBuilderTest.php
```diff
@@ -178,7 +178,7 @@ public function testGetExpressionBuilderException(): void
 
         $this->expectException(Exception::class);
 
-        $expression = new class () implements ExpressionInterface {
+        $expression = new class() implements ExpressionInterface {
```
Tigrov, line 181:
```suggestion
        $expression = new class implements ExpressionInterface {
```

Tigrov (changes requested):

Tigrov:
Added this to the 2.0.0 milestone due to any change of interfaces is BC

src/QueryBuilder/DDLQueryBuilderInterface.php
```diff
@@ -424,4 +424,15 @@ public function renameTable(string $oldName, string $newName): string;
...
+    /**
+     * Refresh materialized view
+     *
+     * @param string $viewName The name of the view to refresh.
+     * @param bool $concurrently Refresh the materialized view without locking out concurrent selects on the materialized view.
+     * @param bool|null $withData When `true` then the backing query is executed to provide the new data. Otherwise if `false` then no new data is generated and the materialized view is left in an unscannable state
+     *
+     * @return string The `REFRESH MATERIALIZED VIEW` SQL statement
+     */
+    public function refreshMaterializedView(string $viewName, bool $concurrently = false, ?bool $withData = null): string;
```
Tigrov, line 437:
```suggestion
    public function refreshMaterializedView(string $viewName, bool $concurrently = false): string;
```

Not sure the argument `$withData` is needed. If it is true, it is equal to "truncate" the materialized view. Maybe beter remove it and if the method is required add `truncateMaterializedView()`

Tigrov (changes requested):

Tigrov (changes requested):

src/Command/CommandInterface.php
smirnov-e: From what I see in [MS SQL](https://learn.microsoft.com/ru-ru/sql/t-sql/statements/alter-materialized-view-transact-sql?view=azure-sqldw-latest) / [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/18/dwhsg/refreshing-materialized-views.html#GUID-EB813C4E-FEE4-4E4C-AAB8-9A34F8EB8D3C) / [Clickhouse](https://clickhouse.com/docs/en/guides/developer/cascading-materialized-views#combining-multiple-source-tables-to-single-target-table) docs:
1. `WITH [NO] DATA` is PostgreSQL only option and pretty much useless one.
2. `REFRESH` exists only on PG and Oracle. MS SQL and Clickhouse do automatic updates, so their implementations should always return `true`.
There are a lot of variants in [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/18/dwhsg/refreshing-materialized-views.html#GUID-51191C38-D52F-4A4D-B6FF-E631965AD69A): 3 commands (`REFRESH`, `REFREASH_ALL_MVIEWS`, `REFRESH_DEPENDENT`) and at least 6 options. Option `out_of_place` seems to be equivalent of CONCURRENTLY in PostgreSQL.

So, the most common part is refresh + option for non-blocking update, but it will be very hard to take full advantage  of database capabilities.

smirnov-e:
> With `create`, `alter`, `drop` and `refresh` methods.

From what I see in [MS SQL](https://learn.microsoft.com/ru-ru/sql/t-sql/statements/alter-materialized-view-transact-sql?view=azure-sqldw-latest) / [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/18/dwhsg/refreshing-materialized-views.html#GUID-EB813C4E-FEE4-4E4C-AAB8-9A34F8EB8D3C) / [Clickhouse](https://clickhouse.com/docs/en/guides/developer/cascading-materialized-views#combining-multiple-source-tables-to-single-target-table) docs:

`DROP`: PostgreSQL requires `DROP MATERIALIZED VIEW`, other databases use simple `DROP VIEW`.

`ALTER` - quite different one from another
* [MS SQL](https://learn.microsoft.com/ru-ru/sql/t-sql/statements/alter-materialized-view-transact-sql?view=azure-sqldw-latest) only allows enable/disable MV
* [PostgreSQL](https://www.postgresql.org/docs/current/sql-altermaterializedview.html) allows modify name / column names / other options excluding source query
* [Clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/alter/view?ysclid=lvtudaqqvx921285858#alter-table--modify-refresh-statement) only allows to change source query  
* [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/18/dwhsg/advanced-materialized-views.html#GUID-D4412AFE-C4BC-4287-A9BB-104AD5E9678E) — refresh mode and type, enable/disable.

`CREATE` — MS SQL, PostgreSQL and Oracle has common syntax part, but Oracle has lots of useful extensions. All except MS SQL have option to skip data population (WITH NO DATA).
* [PostgreSQL](https://www.postgresql.org/docs/current/sql-creatematerializedview.html) — column names, storage options and other.
* [MS SQL](https://learn.microsoft.com/ru-ru/sql/t-sql/statements/create-materialized-view-as-select-transact-sql?view=azure-sqldw-latest) — data distribution 
* [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/18/dwhsg/basic-materialized-views.html#GUID-9C3332BB-53E2-4179-B05C-5C8D1E46EFDF) — additional query for faster refresh, syntax for automatic refresh and partitioning.
* [Clickhouse](https://clickhouse.com/docs/en/sql-reference/statements/create/view#materialized-view) — just query and SQL SECURITY.

Issue #805, opened 2024-02-01
Bug when try to `Command::update()` using `Expression`s in values or in where condition


Tigrov:
There are two problems when update table using expressions:
1. `Expression::$params` are not passed in the query.
2. If `Expression::$params` have an `ExpressionInterface` inside, it is not processed

```php
$values = ['abc'];

$command->update(
    'table_name,
    [
        'values' => new Expression(
            'ARRAY(SELECT DISTINCT UNNEST("values" || :values))',
            ['values' => new ArrayExpression($values, 'varchar(126)')]
        ),
    ],
    '"values" && :values',
    ['values' => new ArrayExpression($values, 'varchar(126)')]
)->execute();
```

### Expected query ###

```sql
UPDATE "table_name" 
SET "values"=ARRAY(SELECT DISTINCT UNNEST("values" || ARRAY[:qp1]::varchar(126)[])) 
WHERE "values" && ARRAY[:qp2]::varchar(126)[]
```

```php
$params === [
    ':qp1' => 'abc',
    ':qp2' => 'abc',
];
```

### Actual query ###

```sql
UPDATE "table_name"
SET "values"=ARRAY(SELECT DISTINCT UNNEST("values" || :values))
WHERE "values" && :values
```

```php
$params === [
    ':values' => ArrayExpression,
];
```

PR #806, opened 2024-02-06
Process expressions inside param values when build a query

Tigrov:
* Allows non-unique param names inside `Expression::$params` (they will be replaced with unique names)
* Process `Expression` instances inside `Expression::$params`

### Related PRs
* https://github.com/yiisoft/db-sqlite/pull/289
* https://github.com/yiisoft/db-mysql/pull/321
* https://github.com/yiisoft/db-pgsql/pull/336
* https://github.com/yiisoft/db-mssql/pull/293
* https://github.com/yiisoft/db-oracle/pull/255

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️/❌
| Breaks BC?    | ❌
| Fixed issues  | #805

### DIFF
--- a/tests/Db/Syntax/SqlParserTest.php
+++ b/tests/Db/Syntax/SqlParserTest.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Tests\Db\Syntax;
+
+use Yiisoft\Db\Tests\AbstractSqlParserTest;
+use Yiisoft\Db\Tests\Support\Stub\SqlParser;
+
+class SqlParserTest extends AbstractSqlParserTest
+{
+    protected function createSqlParser(string $sql): SqlParser
+    {
+        return new SqlParser($sql);
+    }
+}
--- a/tests/Provider/SqlParserProvider.php
+++ b/tests/Provider/SqlParserProvider.php
@@ -93,6 +93,11 @@ class SqlParserProvider
                 ':name',
                 65,
             ],
+            [
+                "SELECT * FROM {{customer}} WHERE \"Ģ:name\" = '😀 ȧ Привет 中文 :label' AND age = :age",
+                ':age',
+                92,
+            ],
         ];
     }
 


src/Expression/ExpressionBuilder.php
```diff
@@ -17,9 +22,99 @@
...
-        return $expression->__toString();
+        $sql = $expression->__toString();
+        $expressionParams = $expression->getParams();
+
+        if (empty($expressionParams)) {
+            return $sql;
+        }
+
+        if ($this->queryBuilder === null || isset($expressionParams[0])) {
+            $params = array_merge($params, $expressionParams);
```
vjik, line 39: Result params can be incorrect, when `$params` use string keys.

src/Expression/ExpressionBuilder.php
```diff
@@ -17,9 +22,99 @@
...
+        $params += $expressionParams;
+
+        if (empty($nonUniqueParams)) {
+            return $sql;
+        }
+
+        $patterns = [];
+        $replacements = [];
+
+        /** @var string $name */
```
vjik, line 60: `$name` can be integer, params may be list (int keys) or array with string keys.

vjik (changes requested):

src/Expression/ExpressionBuilder.php
Tigrov: It is not possible due to `return` on line 40 above
https://github.com/yiisoft/db/pull/806/files#diff-11971291de3a5da6bd65a674be0a96555c044e0dbfa49c18a4a42554edfddf92R40

src/Expression/ExpressionBuilder.php
Tigrov: For this case the behavior is the same as before.

src/Expression/ExpressionBuilder.php
```diff
@@ -17,9 +22,99 @@
...
+            $uniqueName = $this->getUniqueName($name, $params);
+
+            $replacements[] = $uniqueName[0] !== ':' ? ":$uniqueName" : $uniqueName;
+
+            $params[$uniqueName] = $value;
+            $expressionParams[$uniqueName] = $value;
+            unset($expressionParams[$name]);
+        }
+
+        return preg_replace($patterns, $replacements, $sql, 1);
```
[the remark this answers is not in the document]
Tigrov: Fixed

src/Expression/ExpressionBuilder.php
```diff
@@ -14,12 +21,178 @@
  *
  * These expressions can be used with the query builder to build complex and customizable database queries
  * {@see Expression} class.
+ *
+ * @psalm-import-type ParamsType from ConnectionInterface
  */
 class ExpressionBuilder implements ExpressionBuilderInterface
 {
+    public function __construct(private QueryBuilderInterface|null $queryBuilder = null)
```
vjik, line 29: Why is query builder optional?

Tigrov: For BC. In the next maror version `null` can be removed.

src/Syntax/SqlParser.php
```diff
@@ -0,0 +1,183 @@
...
+                ':' => ($word = $this->parseWord()) === ''
+                    ? $this->skipChars(':')
+                    : $result = ':' . $word,
+                '"', "'" => $this->skipQuotedWithoutEscape($this->sql[$pos]),
+                '-' => $this->sql[$this->position] === '-'
+                    ? ++$this->position && $this->skipToAfterChar("\n")
+                    : null,
+                '/' => $this->sql[$this->position] === '*'
+                    ? ++$this->position && $this->skipToAfterString('*/')
+                    : null,
```
vjik, line 60: May be make parser abstract (all current Yii DB implementations has own parser) and make abstract method, that allow implementations of this abstraction add or replace code for symbols?

Tigrov: Thought about this. This is SQL-standart implementation but each DBMS has own quoting features.

Ok, I'll make it abstract.

Tigrov: Done

src/Expression/AbstractExpressionBuilder.php
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @return string[] Replacements for non-unique parameters.
+     */
+    private function appendParams(array &$expressionParams, array &$params): array
+    {
+        $nonUniqueParams = [];
+
+        /** @var non-empty-string $name */
+        foreach ($expressionParams as $name => $value) {
+            $paramName = $name[0] === ':' ? substr($name, 1) : $name;
```
samdark, line 90: Should we support Unicode in parameter names?

src/Syntax/AbstractSqlParser.php
```diff
@@ -0,0 +1,153 @@
...
+     * @var int Current position in SQL statement.
+     */
+    protected int $position = 0;
+
+    /**
+     * @param string $sql SQL statement to parse.
+     */
+    public function __construct(protected string $sql)
+    {
+        $this->length = strlen($sql);
```
samdark, line 32: Could SQL contain unicode strings?

src/Expression/AbstractExpressionBuilder.php
Tigrov: Placeholder must match the pattern `[:][a-zA-Z0-9_]+` and parameter name may or may not contain `:` at the first position.

src/Syntax/AbstractSqlParser.php
Tigrov: Yes, I'll check this.

Tigrov: No problems with UTF-8 or any single-byte encoding.

Should it support other multibyte encodings? If should, it could be a problem to pass that encoding to the parser.

vjik (approved):

samdark (approved):

PR #809, opened 2024-02-12
Add psalm type for parameters to bind to the SQL statement

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


src/Connection/ConnectionInterface.php
```diff
@@ -25,6 +25,8 @@
  *
  * It allows you to access and manipulate databases in a database-agnostic way, so you can write code that works with
  * different database systems without having to worry about the specific details of each one.
+ *
+ * @psalm-type ParamsType = array<string,mixed>|list<mixed>
```
Tigrov, line 29:
```suggestion
 * @psalm-type ParamsType = array<non-empty-string, mixed>|list<mixed>
```
Name of a parameter should be non-empty

Tigrov (approved):

Issue #821, opened 2024-04-16
Needs to refactor `Query::each()`


Tigrov:
`Query::each()` returns data reader to get data row by row.

```php
foreach ($query->each() as $row) {}
```

Current realization of the data reader reads batch of rows then read each row from the rows.

https://github.com/yiisoft/db/blob/e380ae2ffb6f3a466e6f168fafa93548e50f0d8c/src/Query/BatchQueryResult.php#L70-L85

If read the result of query row by row (instead of batch of rows), this will be more efficient and will use less memory.

Issue #716, opened 2023-05-04
Non-consistent types in `BatchQueryResultInterface`


vjik:
Interface `BatchQueryResultInterface` [extends](https://github.com/yiisoft/db/blob/2893abddc82dc1f89dec0e36e9c7beeb1a0a4ad1/src/Query/BatchQueryResultInterface.php#L37) `Iterator<int|string, mixed>`, but [method](https://github.com/yiisoft/db/blob/master/src/Query/BatchQueryResultInterface.php#LL77C1-L77C1) `key()` returns `int|string|null`

PR #822, opened 2024-04-17
Refactor data readers

Tigrov:
* Refactor `DataReader` class
* Rename `DataReader` to `PdoDataReader` class
* Refactor `BatchQueryResult`
* Use `DataReaderInterface` as a result of method `Query::each()`

As a result `->rewind()` can be called multiple times before iteration. It allows to get access to the first row before iteration.

### Related PRs
- yiisoft/db-mssql#300
- https://github.com/yiisoft/active-record/pull/409

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #821, #716


src/Query/DataReaderInterface.php
```diff
@@ -0,0 +1,69 @@
...
+ * }
+ *
+ * Note: That since DataReader is a forward-only stream, you can only traverse it once. Doing it the second time will
+ * throw an exception.
+ *
+ * @extends Iterator<int|string|null, array|false>
+ *
+ * @psalm-import-type IndexBy from QueryPartsInterface
+ */
+interface DataReaderInterface extends Iterator, Countable
```
samdark, line 33: Consider reusing interfaces from https://github.com/yiisoft/data/blob/master/src/Reader/DataReaderInterface.php. Not mandatory though...

samdark (approved):

src/Query/DataReaderInterface.php
Tigrov: They have the same name but different meaning
- `Data\Reader\DataReaderInterface` is for reading from `QueryInterface`
- `Db\Query\DataReaderInterface` is for reading from `PDOStatement`

vjik (approved):

vjik (approved):

samdark (approved):

PR #832, opened 2024-05-04
Fix psalm

Tigrov:
1. Replace `paths-ignore` to clear `paths`
2. Fix issue with `MissingClassConstType` for PHP version less then `8.3`
3. Added two new parameters to config:
    * findUnusedVariablesAndParams="true"
    * findUnusedPsalmSuppress="true"
4. Clear code

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "rector/rector": "^1.0",
         "roave/infection-static-analysis-plugin": "^1.16",
         "spatie/phpunit-watcher": "^1.23",
-        "vimeo/psalm": "^4.30|^5.20",
+        "vimeo/psalm": "^5.24",
         "yiisoft/aliases": "^3.0",
         "yiisoft/cache-file": "^3.1",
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "rector/rector": "^1.0",
         "roave/infection-static-analysis-plugin": "^1.16",
         "spatie/phpunit-watcher": "^1.23",
-        "vimeo/psalm": "^4.30|^5.20",
+        "vimeo/psalm": "^4.30|^5.24",
```
vjik, line 37:
```suggestion
        "vimeo/psalm": "^5.24",
```

vjik (approved):

composer.json
Tigrov: Tried this way (the first commit). It is cause of errors in db drivers when composer install.
Will be changed after PHP version is raised to 8.1 for all db drivers.

PR #855, opened 2024-06-29
Add array and json overlaps conditions

Tigrov:
### Related PRs
- yiisoft/db-mysql#342
- yiisoft/db-sqlite#310
- yiisoft/db-pgsql#350

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | yiisoft/active-record#362


src/QueryBuilder/Condition/JsonOverlapsCondition.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\QueryBuilder\Condition;
+
+/**
+ * Condition that represents `JSON OVERLAPS` operator and is used to check if a column of json type overlaps an array.
```
samdark, line 8:
```suggestion
 * Condition that represents `JSON OVERLAPS` operator and is used to check if a column of JSON type overlaps an array.
```

samdark (changes requested):

samdark (approved):

Issue #872, opened 2024-08-27
Add SensitiveParameter attribute


samdark:
https://github.com/yiisoft/db/blob/d1da00342a92e896f77e074ee039a711f1edb6a2/src/Driver/Pdo/PdoDriverInterface.php#L81

Add `#[\SensitiveParameter]` to password.

heap-s:
Tried to fix but haven't contributed much in the past. Not so sure why checks are failing, if you could give some guidance I'll raise the PR again, if not, no worries

samdark:
@heap-s it does't look like it's your fault. Something's broken in our tests pipeline. Pull request itself looked OK.

heap-s:
@samdark I'll reopen today then to resolve this!

PR #874, opened 2024-08-28
Pdo driver interface/password sensitive param

heap-s:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #872

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 - Enh #862: Refactor PHP type of `ColumnSchemaInterface` instances (@Tigrov)
 - Enh #865: Raise minimum PHP version to `^8.1` with minor refactoring (@Tigrov, @vjik)
 - Enh #798: Allow `QueryInterface::one()` and `QueryInterface::all()` to return objects (@darkdef, @Tigrov)
-- Chg #872: Add `#[\SensitiveParameter]` attribute for password (@heap-s)
+- Enh #872: Use `#[\SensitiveParameter]` attribute to mark sensitive parameters (@heap-s)
 
 ## 1.3.0 March 21, 2024
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@
 - Enh #865: Raise minimum PHP version to `^8.1` with minor refactoring (@Tigrov, @vjik)
 - Enh #798: Allow `QueryInterface::one()` and `QueryInterface::all()` to return objects (@darkdef, @Tigrov)
 - Enh #872: Use `#[\SensitiveParameter]` attribute to mark sensitive parameters (@heap-s)
+- Enh #864: Realize column factory (@Tigrov)
+- Enh #875: Ignore "Packets out of order..." warnings in `AbstractPdoCommand::internalExecute()` method (@Tigrov)
 
 ## 1.3.0 March 21, 2024
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Enh #862: Refactor PHP type of `ColumnSchemaInterface` instances (@Tigrov)
 - Enh #865: Raise minimum PHP version to `^8.1` with minor refactoring (@Tigrov, @vjik)
 - Enh #798: Allow `QueryInterface::one()` and `QueryInterface::all()` to return objects (@darkdef, @Tigrov)
+- Chg #872: Add `#[\SensitiveParameter]` attribute for password (@heap-s)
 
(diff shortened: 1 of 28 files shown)


src/Driver/Pdo/PdoDriverInterface.php
```diff
@@ -78,7 +78,7 @@ public function getUsername(): string;
      *
      * @param string $password The password for establishing DB connection.
      */
-    public function password(string $password): void;
+    public function password(#[\SensitiveParameter] string $password): void;
```
vjik, line 81: Attribute not need in interface. Attribute should be added in implementation. Example: https://3v4l.org/KNALl

vjik (changes requested):

src/Driver/Pdo/PdoDriverInterface.php
heap-s: So you'd prefer to see this in AbstractPdoDriver class that implements PdoDriverInterface?

vjik: Yes

heap-s: I believe this is resolved, thanks for the help!

vjik: Add to constructor parameter also. And need line in changelog.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 - Enh #862: Refactor PHP type of `ColumnSchemaInterface` instances (@Tigrov)
 - Enh #865: Raise minimum PHP version to `^8.1` with minor refactoring (@Tigrov, @vjik)
 - Enh #798: Allow `QueryInterface::one()` and `QueryInterface::all()` to return objects (@darkdef, @Tigrov)
+- Chg #872: Add `#[\SensitiveParameter]` attribute for password (@heap-s)
```
vjik, line 37:
```suggestion
- Enh #872: Use `#[\SensitiveParameter]` attribute to mark sensitive parameters (@heap-s)
```

vjik (approved):

CHANGELOG.md
heap-s: I believe that I've made all of your suggested changes, sorry this has been quite the process. Now that I understand the process for contributing a bit better here, I can make this easier in the future.

samdark (approved):

CHANGELOG.md
vjik: It's OK. Thanks for contribution 👍

Tigrov:
Can we test it?

Does it make sence to add it to the interface?

https://github.com/yiisoft/db/blob/99d69c17679dd306059c19797bd587b70582a875/src/Driver/Pdo/PdoDriverInterface.php#L81

samdark:
@Tigrov not really.

vjik:
Hm, if we can throw exception, then we can check trace. Example: https://3v4l.org/OYZAt

Tigrov:
To throw an exception we can pass an integer value instead of string

heap-s:
Have we decided what we're doing here in terms of testing?

Tigrov:
> Have we decided what we're doing here in terms of testing?

Can you try to use this example for testing the SensitiveParameter?

https://3v4l.org/Gh1lY

heap-s:
@Tigrov The solution that I think of implementing is https://3v4l.org/QK6Yo I believe that this is appropriate and what you'd like in terms of testing for the SensitiveParameter. Let me know what you think.

Tigrov:
> Let me know if you'd like me to squash these commits, so that the history is cleaner

Let's see on test results then squash and merge it.

vjik (approved):

tests/Db/Driver/PDO/PDODriverTest.php
```diff
@@ -62,4 +62,23 @@ public function testGetUsername(): void
...
+            new PDODriver($dsn, password: null);
+        } catch (\TypeError $e) {
+            $this->assertTrue($e->getTrace()[0]['args'][2] instanceof \SensitiveParameterValue);
+        }
+        $pdoDriver = new PDODriver($dsn);
+        try {
+            $pdoDriver->password(null);
+        } catch (\TypeError $e) {
+            $this->assertTrue($e->getTrace()[0]['args'][0] instanceof \SensitiveParameterValue);
+        }
```
Tigrov, line 82:
```suggestion
        }
```

Tigrov (approved):

Tigrov:
Thanks 👍

Issue #854, opened 2024-06-22
New concept of columns for version 2.0


Tigrov:
Main changes:

- [ ] 1. Remove `ColumnSchema::$name` and `ColumnSchema::$precision` (use `ColumnSchema::$size`) **(under discussion)**
- [ ] 2. Rename `ColumnSchema::$enumValues` to `ColumnSchema::$values` (related with some issues) **(under discussion)**
- [ ] 3. Add `ColumnSchema::getFullDbType()` method to get full name of the type, e.g. `varchar(20)` #695
- [ ] 4. Remove `pk` abstract type constants `ColumnSchemaInterface::...pk` use `ColumnBuilder::...pk()` instead yiisoft/db-migration#6
- [ ] 5. Rename `AbstractColumnSchema` to `Column`
- [ ] 6. Realize `TypeColumn` classes (`IntegerColumn`, `StringColumn`, etc) #752
- [ ] 7. Realize `ColumnFactory`, `ColumnBuilder` and `ColumnDefinitionBuilder`
- [ ] 8. Add constructor to `TableSchema` (instead of `AbstractSchema::resolveTableName()`)

The changes will solve few issues and after related improvements it will solve few more issues.
It will allow specify any php type or object type for any db type or table column.

There is a draft #808

PR #882, opened 2024-10-07
Move `ArrayColumnSchema` and `StructuredColumnSchema` from `db-pgsql`

Tigrov:
### Related PR
- https://github.com/yiisoft/db-pgsql/pull/363

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️/❌
| Fixed issues  | #854


src/Expression/StructuredExpression.php
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Will be encoded to `ROW(10, USD)` in PostgreSQL.
+ */
+final class StructuredExpression implements ExpressionInterface
+{
+    /**
+     * @param ColumnSchemaInterface[] $columns
+     * @psalm-param array<string, ColumnSchemaInterface> $columns
+     */
+    public function __construct(
```
samdark, line 34: Need phpdoc here badly. It's not obvious about what to pass where. Duplicating docs from getters is alright for this case.

src/Expression/StructuredExpression.php
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @return ColumnSchemaInterface[]
+     */
+    public function getColumns(): array
+    {
+        return $this->columns;
+    }
+
+    /**
+     * The content of the structured type. It can be represented as an associative array of structured type column names
```
samdark, line 65: How else can it be represented? Why `mixed`?

src/Expression/StructuredExpression.php
```diff
@@ -0,0 +1,103 @@
...
+     * and values.
+     */
+    public function getValue(): mixed
+    {
+        return $this->value;
+    }
+
+    /**
+     * Sorted values according to the order of structured type columns,
+     * indexed keys are replaced with column names,
```
samdark, line 75: Does that mean that it's always an array?

samdark (approved):

src/Expression/StructuredExpression.php
Tigrov: Done

src/Expression/StructuredExpression.php
Tigrov: Replaced `mixed` to `array|object` and added description about possible values

src/Expression/StructuredExpression.php
Tigrov: If columns are specified the value will be converted to an array with normalization, otherwise returns value as is. Expressions also returned as is.

Issue #851, opened 2024-05-29
In `QueryInterface::scalar()` result no different between value `false` and no results


vjik:

samdark:
Can't differentiate `false` as "nothing" from `false` as value.

Tigrov:
We can mention in the documentation do not use the method for `bool` results.

samdark:
Should be added to phpdoc, yes.

PR #918, opened 2025-01-09
Add PHPDoc about `boolean` values in `scalar()` and `queryScalar()`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #851

### DIFF
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -680,8 +680,8 @@ interface CommandInterface
 
     /**
      * Execute the SQL statement and returns the value of the first column in the first row of data.
-     *
      * This method is best used when you need only a single value.
+     * Do not use this method for `boolean` values as it returns `false` if there is no value.
      *
      * @throws Exception
      * @throws Throwable If execution failed.
--- a/src/Query/QueryInterface.php
+++ b/src/Query/QueryInterface.php
@@ -288,8 +288,8 @@ interface QueryInterface extends ExpressionInterface, QueryPartsInterface, Query
 
     /**
      * Returns the query results as a scalar value.
-     *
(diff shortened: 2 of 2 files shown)


vjik (commented):
Suggest to add `scalar $default = null` parameter. This allow to use this methods for boolean values also.

Tigrov:
How to use it?

Field of boolean type can return `true`, `false` and `null` values.

samdark (approved):

vjik:
> How to use it?
> 
> Field of boolean type can return `true`, `false` and `null` values.

When I know that result `bool|null`, I can use `->queryScalar('')` code. When returns empty string - value doesn't exist.

Tigrov:
Good idea, but it uses `PDOStatement::fetchColumn()` and the function returns `false` for both: no results and `false` value result
https://github.com/yiisoft/db/blob/98eed3919245d76ad2123c60692c854616ea6f01/src/Driver/Pdo/AbstractPdoCommand.php#L251-L253

From the documentation https://www.php.net/manual/en/pdostatement.fetchcolumn.php

```
Note:
PDOStatement::fetchColumn() should not be used to retrieve boolean columns, as it is impossible to distinguish a value of [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) from there being no more rows to retrieve. Use [PDOStatement::fetch()](https://www.php.net/manual/en/pdostatement.fetch.php) instead.
```

vjik:
> Good idea, but it uses `PDOStatement::fetchColumn()` and the method returns `false` for both: no results and `false` value result

So in this case my suggestion doesn't matter.

vjik (approved):

PR #920, opened 2025-01-12
Move index type constants to `IndexType` class

Tigrov:
### Related PRs
- https://github.com/yiisoft/db-mysql/pull/374
- https://github.com/yiisoft/db-pgsql/pull/384
- https://github.com/yiisoft/db-mssql/pull/339
- https://github.com/yiisoft/db-oracle/pull/301
- https://github.com/yiisoft/db-migration/pull/285

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Constant/IndexType.php
+++ b/src/Constant/IndexType.php
@@ -5,74 +5,13 @@ declare(strict_types=1);
 namespace Yiisoft\Db\Constant;
 
 /**
- * Defines the available index types. It is used in {@see DDLQueryBuilderInterface::createIndex()}.
+ * Defines the available index types for {@see DDLQueryBuilderInterface::createIndex()} method.
+ * Use driver specific implementations for other supported types if any.
  */
 final class IndexType
 {
     /**
      * Define the type of the index as `UNIQUE`.
-     *
-     * Supported by `MySQL`, `MariaDB`, `MSSQL`, `Oracle`, `PostgreSQL`, `SQLite`.
      */
     public const UNIQUE = 'UNIQUE';
-    /**
-     * Define the type of the index as `BTREE`.
-     *
-     * Supported by `MySQL`, `PostgreSQL`.
-     */
-    public const BTREE = 'BTREE';
-    /**
-     * Define the type of the index as `HASH`.
-     *
-     * Supported by `MySQL`, `PostgreSQL`.
-     */
-    public const HASH = 'HASH';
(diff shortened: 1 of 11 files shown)


src/Constant/IndexType.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Constant;
+
+/**
+ * Defines the available index types. It is used in {@see DDLQueryBuilderInterface::createIndex()}.
+ */
+final class IndexType
```
vjik, line 10: Each DBMS has own list of index types, which almost do not overlap. May be remove `IndexType` from Yii DB and create separate `IndexType` in each implementation package?

Tigrov: Maybe just remove it. As far as I see only `UNIQUE` keyword makes sense.

**DBMS** | **Index types**
-------|------------
**MYSQL** | UNIQUE, FULLTEXT, SPATIAL
**PGSQL** | UNIQUE
**MSSQL** | UNIQUE, UNIQUE CLUSTERED, CLUSTERED, SPATIAL, COLUMNSTORE, CLUSTERED COLUMNSTORE, XML, PRIMARY XML
**SQLITE** | UNIQUE
**ORACLE** | UNIQUE, BITMAP, MULTIVALUE

**DBMS** | **Index methods**
-----------|------------------
**MYSQL** | BTREE (by default), HASH, RTREE (MARIADB)
**PGSQL** | BTREE (by default), HASH, GIST, SPGIST, GIN, BRIN
**MSSQL** | BTREE (there is no option)
**SQLITE** | BTREE (there is no option)
**ORACLE** | BTREE (there is no option), depends on index type

vjik: Also other indexes can potentially be added using extensions. It seems, there is such opportunity in PostgreSQL.

Сlasses with constants may be convient to use in userland. Remove it from Yii DB, but add for DBMS-specific packages. WDYT?

samdark (approved):

src/Constant/IndexType.php
Tigrov: Yes, it looks good solution.

vjik (approved):

PR #925, opened 2025-01-27
Add callback to `Query::all()` and `Query::one()`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | https://github.com/yiisoft/data-db/issues/50


UPGRADE.md
```diff
@@ -115,6 +115,8 @@ Each table column has its own class in the `Yiisoft\Db\Schema\Column` namespace
 ### New methods
 
 - `QuoterInterface::getRawTableName()` - returns the raw table name without quotes;
+- `QueryInterface::callback()` - allows to use a callback, which should be called on each row of the query result;
+- `QueryInterface::getCallback()` - returns the callback to be called on each row of the query result or `null` if not set;
```
samdark, line 119:
```suggestion
- `QueryInterface::getCallback()` - returns the callback to be called on each row of the query result or `null` if the callback is not set;
```

src/Query/Query.php
```diff
@@ -250,6 +272,12 @@ public function batch(int $batchSize = 100): BatchQueryResultInterface
         ;
     }
 
+    public function callback(Closure|null $callback): static
```
samdark, line 275: Need phpdoc with the callback signature.

src/Query/QueryInterface.php
```diff
@@ -81,6 +82,13 @@ public function all(): array;
      */
     public function batch(int $batchSize = 100): BatchQueryResultInterface;
 
+    /**
+     * Sets callback, which should be called on each row of the query result.
+     *
+     * @psalm-param CallbackType|null $callback
```
samdark, line 88: Need to mention callback signature.

samdark (changes requested):

src/Query/QueryInterface.php
```diff
@@ -30,6 +30,7 @@
  * @psalm-type IndexBy = Closure(array):array-key|string
  * @psalm-import-type ParamsType from ConnectionInterface
  * @psalm-import-type SelectValue from QueryPartsInterface
+ * @psalm-type ResultCallback = Closure(non-empty-array<array>):non-empty-array<array|object>
```
vjik, line 33:
```suggestion
 * @psalm-type ResultCallback = Closure(non-empty-array<array|object>):non-empty-array<array|object>
```

vjik (changes requested):

vjik:
Add line to changelog

src/Query/QueryInterface.php
Tigrov: We get only arrays inside query.

samdark (approved):

src/Query/QueryInterface.php
vjik: OK

vjik (approved):

PR #939, opened 2025-03-13
Case sensitive option in `LikeCondition`

### LINKED ISSUE #887: Add case sensitive option to like condition

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
Fix #887 

- https://github.com/yiisoft/db-sqlite/pull/344
- https://github.com/yiisoft/db-pgsql/pull/391
- https://github.com/yiisoft/db-mysql/pull/383
- https://github.com/yiisoft/db-mssql/pull/349
- https://github.com/yiisoft/db-oracle/pull/311


src/QueryBuilder/Condition/Builder/LikeConditionBuilder.php
```diff
@@ -70,39 +68,59 @@ public function build(LikeConditionInterface $expression, array &$params = []):
...
+        string|ExpressionInterface $value,
+        LikeConditionInterface $expression,
+        array|null $escape,
+        array &$params,
+    ): string {
+        if ($value instanceof ExpressionInterface) {
+            return $this->queryBuilder->buildExpression($value, $params);
+        }
+        return $this->queryBuilder->bindParam(
+            $escape === null ? $value : ('%' . strtr($value, $escape) . '%'),
```
Tigrov, line 109:
```suggestion
            new Param($escape === null ? $value : ('%' . strtr($value, $escape) . '%'), DataType::STRING),
```

vjik: done

tests/AbstractQueryBuilderTest.php
```diff
@@ -476,7 +462,16 @@ public function testBuildLikeCondition(
...
-        $this->assertSame($expectedParams, $params);
+        $this->assertSame(array_keys($expectedParams), array_keys($params));
+        foreach ($params as $name => $value) {
+            if ($value instanceof Param) {
+                $this->assertInstanceOf(Param::class, $expectedParams[$name]);
+                $this->assertSame($expectedParams[$name]->getValue(), $value->getValue());
+                $this->assertSame($expectedParams[$name]->getType(), $value->getType());
+            } else {
+                $this->assertSame($expectedParams[$name], $value);
+            }
```
Tigrov, line 473:
```suggestion
            $this->assertEquals($expectedParams[$name], $value);
```
Could be replaced with `assertEquals()`

Tigrov (approved):

tests/AbstractQueryBuilderTest.php
vjik: `assertEquals` don't check types. Integer `3` equals to string `"3"`. In this case we should check parameter value type also.

Tigrov: In our case it is always string, but if it is necessary - ok

Issue #771, opened 2023-08-17
Support Query result typing


smirnov-e:
Currently 
- data returned by ActiveQuery is typed if column name matches table column name. 
- data returned by Query is mostly untyped.

```sql
CREATE TABLE test (
     id INTEGER GENERATED ALWAYS AS IDENTITY,
     int INTEGER,
     float FLOAT,
     int_array INT[],
     text_array TEXT[]
 );

CREATE TYPE currency_money AS
(
  value numeric(19,4),
  currency_code char(3)
);
 
 INSERT INTO test 
 ("int", "float", int_array, text_array)
VALUES (1, 2.33, ARRAY[1, 2, 3], ARRAY['test']);
```
Code like this 
```php
$db->createCommand('SELECT * FROM test')->queryAll();
// or like this

$db->createCommand('SELECT 1 AS id, 
1 AS int, 
2.3 AS float, 
ARRAY[1,2] AS int_array, 
ARRAY['test'] AS text_array,
ARRAY['(0,USD)'::currency_money] as money_array
')->queryAll();
```
will return 
```php
[
    [
        'id' => 1,
        'int' => 1,
        'float' => '2.33', // type lost
        'int_array' => '{1,2,3}', // type lost
        'text_array' => '{test}', // type lost
        'money_array' => '{"(0.0000,USD)"}', // type lost
    ],
]
```
So currently in Yii3 and Yii2 to get typed data you have to extensively map values and manually control returned types.

```php 
// like this
$price = (new Query($db))->select(['price'])->from('product')->scalar();
$pricePhpType = $tableSchema->getColumn('price')->phpTypecast($price);
// or like this
array_map($rows, function(array $row) {
     $row['int_array'] = (new ArrayParser())->parse($row['int_array']));
     return $row;
})
```

Other way is to use `PDOStatement::getColumnMeta` which returns rather fine type information:
```
ARRAY(
    [pgsql:oid] => 16413
    [pgsql:table_oid] => 0
    [native_type] => _currency_money
    [pdo_type] => 2
    [name] => money_array
    [len] => -1
    [precision] => -1
)
```
```
[native_type] => int4 // int 
[native_type] => numeric // int 
[native_type] => _int4 // array of INT.  "_<typename>" means array of <typename>
[native_type] => _text // array of strings
[native_type] => _currency_money // array of currency_money (custom type)
```
[...]

PR #941, opened 2025-03-18
Implement the ability for user-defined type casting

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/docs/guide/en/schema/typecasting.md
+++ b/docs/guide/en/schema/typecasting.md
@@ -1,0 +1,281 @@
+# Type casting values
+
+Type casting is the process of converting a value from one data type to another. In the context of the database,
+type casting is used to ensure that values are saved and retrieved in the correct type.
+
+## Casting values to be saved in the database
+
+When saving a value to the database, the value must be in the correct type. For example, if saving a value to a column
+that is of type `integer`, the value must be an integer.
+
+To ensure that the value is saved in the correct type, `ColumnInterface::dbTypecast()` method can be used to cast 
+the value to the correct type. Some methods in the DB library, such as `CommandInterface::insert()`, automatically
+cast values to the correct type.
+
+```php
+use Yiisoft\Db\Connection\ConnectionInterface;
+
+/** @var ConnectionInterface $db */
+$command = $db->createCommand();
+$command->insert('customer', [
+    'name' => 'John Doe',
+    'is_active' => true,
+]);
+$command->execute();
+```
+
+In the example above, the value of `is_active` is a `boolean`, but the column `is_active` can be of type `bit`.
+The `CommandInterface::insert()` method will automatically cast the value to the correct type.
+
+## Casting values retrieved from the database
+
+When you retrieve a value from the database, the value can be returned in a different type than you expect.
+For example, a value that is stored as an `float` in the database can be returned as a `string`. This is because 
+the database driver do not convert some data types when retrieves values.
+
+To ensure that the value is returned in the correct type, you can use `ColumnInterface::phpTypecast()` method to cast 
+the value to the correct type.
+
+```php
+use Yiisoft\Db\Connection\ConnectionInterface;
+
+/** @var ConnectionInterface $db */
+$command = $db->createCommand('SELECT * FROM {{customer}} WHERE id = 1');
+
+$row = $command->queryOne();
+$isActive = $row['is_active'];
+
+// Cast the value to the correct type
+$isActive = $db->getTableSchema('customer')->getColumn('is_active')->phpTypecast($isActive);
+```
+
+In the example above, the value of `is_active` can be retrieved from the database as a `bit`, but the correct PHP type 
+is `boolean`. The `ColumnInterface::phpTypecast()` method is used to cast the value to the correct type.
+
+## Custom type casting
+
+To implement custom type casting it needs to extend the `AbstractColumnSchema` class and overriding the `dbTypecast()` 
+and `phpTypecast()` methods.
+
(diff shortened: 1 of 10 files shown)


docs/guide/en/schema/typecasting.md
```diff
@@ -0,0 +1,281 @@
...
+use Yiisoft\Db\Connection\ConnectionInterface;
+
+/** @var ConnectionInterface $db */
+$command = $db->createCommand('SELECT * FROM {{customer}} WHERE id = 1');
+
+$row = $command->queryOne();
+$isActive = $row['is_active'];
+
+// Cast the value to the correct type
+$isActive = $db->getTableSchema('customer')->getColumn('is_active')->phpTypecast($isActive);
```
samdark, line 49: Is there a way to cast the whole row at once?

docs/guide/en/schema/typecasting.md
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    public function getY(): float
+    {
+        return $this->y;
+    }
+}
+```
+
+Then it needs to use the custom column class in the database connection configuration.
```
samdark, line 122:
```suggestion
Then use the custom column class in the database connection configuration.
```

docs/guide/en/schema/typecasting.md
```diff
@@ -0,0 +1,281 @@
...
+
+foreach ($tags as $tag) {
+    echo $tag;
+}
+```
+
+## Structured data types
+
+Some databases support structured data types, such as `composite` types in Postgres. To cast a structured data type to
+a custom class, it needs to create a column class which extends `AbstractStructuredColumn` and override
```
samdark, line 185:
```suggestion
a custom class, you need to create a column class which extends `AbstractStructuredColumn` and override 
```

docs/guide/en/schema/typecasting.md
```diff
@@ -0,0 +1,281 @@
...
+    {
+        return new ArrayIterator([
+            'value' => $this->value,
+            'currency_code' => $this->currencyCode,
+        ]);
+    } 
+}
+```
+
+Then it needs to use the column class in the database connection configuration.
```
samdark, line 266:
```suggestion
Then use the column class in the database connection configuration.
```

samdark (changes requested):

samdark (approved):

docs/guide/en/schema/typecasting.md
Tigrov: No, but there is a task about how to do it automatically
https://github.com/yiisoft/db/issues/771

docs/guide/en/schema/typecasting.md
```diff
@@ -0,0 +1,293 @@
...
+
+    public function getValue(): float
+    {
+        return $this->value;
+    }
+
+    /**
+     * `JsonSerializable` interface is implemented to convert the object to a database representation.
+     */
+    public function jsonSerialize(): array
```
vjik, line 257: May be to add another example with `dbTypecast()` . This allow to keep DTO clear, without infrastructure things.

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -138,13 +182,14 @@ protected function columnDefinitionParser(): ColumnDefinitionParser
...
+     * @psalm-param ColumnInfo $info
+     * @psalm-assert ColumnInfo $info
      * @psalm-return class-string<ColumnInterface>
      */
-    protected function getColumnClass(string $type, array $info = []): string
+    protected function getColumnClass(string $type, array &$info = []): string
     {
-        return match ($type) {
+        /** @psalm-var class-string<ColumnInterface> */
+        return $this->mapType($this->columnClassMap, $type, $info) ?? match ($type) {
```
vjik, line 192: Extract `match ($type) { ...` to separate method. This allow to prevent duplicate `$this->mapType(...)` in `yiisoft/db-*` packages.

docs/guide/en/schema/typecasting.md
Tigrov: This is how to convert object to array, then the array is typecasting and converting to string representation.

Also possible way to declare the properties as public.

https://github.com/yiisoft/db/blob/adf61039ab6dce56cfc08a2c1c3818df387297b5/src/Helper/DbArrayHelper.php#L235-L251

src/Driver/Pdo/AbstractPdoConnection.php
```diff
@@ -50,8 +51,11 @@ abstract class AbstractPdoConnection extends AbstractConnection implements PdoCo
     protected QuoterInterface|null $quoter = null;
     protected SchemaInterface|null $schema = null;
 
-    public function __construct(protected PdoDriverInterface $driver, protected SchemaCache $schemaCache)
-    {
+    public function __construct(
+        protected PdoDriverInterface $driver,
+        protected SchemaCache $schemaCache,
+        protected ColumnFactoryInterface|null $columnFactory = null,
```
vjik, line 57: This must be used in DB implementations, right? Currently is not used.

vjik (changes requested):

src/Driver/Pdo/AbstractPdoConnection.php
Tigrov: This is an option for user defined column factory
https://github.com/yiisoft/db/pull/941/files#diff-05f0a5783bcd92ea3d34bbe0c21bb3b2407837f6aa1efd441c87e5e08436dce3R292

Issue #714, opened 2023-04-24
Remove methods replacing query condition


samdark:
https://github.com/yiisoft/db/blob/master/src/Query/QueryPartsInterface.php#L642

Methods like where() are often the source of problems cause sometimes these are used instead of andWhere(), orWhere().

Possible solutions:

1. Remove such methods. Add reset*() for the rare cases it is needed. 
2. Make where() the same as andWhere(). 
Add reset*() for the rare cases it is needed.

Tigrov:
Use case for the current implementation of `where()`
```php
$query = (new Query($db))->from('table_name');

foreach ($list as $item) {
    $count = $query->where(['field' => $item])->count();
    ...
}
```

For the issues, renaming it to `withWhere()` would be enough
```php
$count = $query->withWhere(['field' => $item])->count();
```

For me the current implementation of `where()` looks better. Perhaps this is a documentation issue for cases where need to use `andWhere()` or `orWhere()` instead of `where()`.

vjik:
_1. Remove such methods. Add reset*() for the rare cases it is needed._

It is will be counter intuitive: 

```php
$query->whe // ???
// Why IDE don't suggest method where?
// Why not exist method where???
// Oh, need `andWhere`
```

Also it's poorly readable:

```php
$query->andWhere(...)->all();
```

Also this suggestion will lead to inconsistency with methods `orderBy()`, `select()`, and so on. Change these methods will leed to more poorly readable.

_2. Make where() the same as andWhere(). Add reset*() for the rare cases it is needed._

This suggesion is better, but in this case need change behavior of methods `orderBy()`, `select()`, and so on also.

Also `reset*()` methods improve code clarity:

```php
$query = (new Query($db))->from('table_name');

// current
foreach ($list as $item) {
    $count = $query->where(['field' => $item])->count();
    ...
}

// suggestion
foreach ($list as $item) {
    $count = $query->resetWhere()->where(['field' => $item])->count();
    ...
}
```

I like this way. despite it is break of BC. 

If cases when methods `where()`, `select()`, etc. used for override previous values is rare, we can definitely do it.

---

@Tigrov Renaming it to `withWhere()` don't resolve issue, because problem in usage exactly `where()` method which is used several times, thinking that conditionals are accumulating.

Tigrov:
The second suggestion looks good.

```php
foreach ($list as $item) {
    $count = $query->resetWhere()->where(['field' => $item])->count();
    ...
}
```

PR #944, opened 2025-03-26
Added `->setWhere()` as force method `->where()`

lav45:
Added `->where(force: true)` to protect against accidental overwriting `where` condition

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | 

```php
$query = (new Query($db))
    ->where(['like', 'name', 'foo%'])
    ->where(['not like', 'name', 'foo%']); // => throw new LogicException()
```

This solves the problem of accidentally overriding the `where` condition.
If you really want to redefine `where`, then you should use the `setWhere()` or `andWhere()` or `orWhere()` method.

```php
$query = (new Query($db))
    ->where(['like', 'name', 'foo%'])
    ->setWhere(['not like', 'name', 'foo%']);

$this->assertEquals(['not like', 'name', 'foo%'], $query->getWhere());
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,6 @@
 - New #942: Allow PHP backed enums as values (@Tigrov)
 - Enh #943: Add `getCacheKey()` and `getCacheTag()` methods to `AbstractPdoSchema` class (@Tigrov)
 - Enh #944: Added `setWhere()` as method a forced for overwriting `where()` (@lav45)
-- Enh #925: Add callback to `Query::all()` and `Query::one()` methods (@Tigrov)
 - Enh #925, #951: Add callback to `Query::all()` and `Query::one()` methods (@Tigrov, @vjik)
 - New #954: Add `DbArrayHelper::arrange()` method (@Tigrov)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@
 - New #939: Add `caseSensitive` option to like condition (@vjik)
 - New #942: Allow PHP backed enums as values (@Tigrov)
 - Enh #943: Add `getCacheKey()` and `getCacheTag()` methods to `AbstractPdoSchema` class (@Tigrov)
+- Enh #944: Added `setWhere()` as method a forced for overwriting `where()` (@lav45)
 
 ## 1.3.0 March 21, 2024
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -120,6 +120,7 @@ Each table column has its own class in the `Yiisoft\Db\Schema\Column` namespace 
 - `QueryInterface::resultCallback()` - allows to use a callback, to be called on all rows of the query result;
 - `QueryInterface::getResultCallback()` - returns the callback to be called on all rows of the query result or 
   `null` if the callback is not set;
+- `QueryPartsInterface::setWhere()` - Overwrites the `WHERE` part of the query;
 - `ConnectionInterface::getColumnFactory()` - returns the column factory object for concrete DBMS;
 - `ConnectionInterface::getServerInfo()` - returns `ServerInfoInterface` instance which provides server information;
 - `QueryBuilderInterface::buildColumnDefinition()` - builds column definition for `CREATE TABLE` statement;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -120,7 +120,7 @@ Each table column has its own class in the `Yiisoft\Db\Schema\Column` namespace 
 - `QueryInterface::resultCallback()` - allows to use a callback, to be called on all rows of the query result;
 - `QueryInterface::getResultCallback()` - returns the callback to be called on all rows of the query result or 
   `null` if the callback is not set;
-- `QueryPartsInterface::setWhere()` - Overwrites the `WHERE` part of the query;
+- `QueryPartsInterface::setWhere()` - overwrites the `WHERE` part of the query;
 - `ConnectionInterface::getColumnFactory()` - returns the column factory object for concrete DBMS;
 - `ConnectionInterface::getServerInfo()` - returns `ServerInfoInterface` instance which provides server information;
 - `QueryBuilderInterface::buildColumnDefinition()` - builds column definition for `CREATE TABLE` statement;
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -699,7 +699,7 @@ class Query implements QueryInterface
         if ($this->where === null) {
             $this->where = $condition;
         } else {
-            throw new LogicException('The `where` condition was set earlier. If you want to overwrite it, use the `setWhere()`, `andWhere()` or `orWhere()` method.');
+            throw new LogicException('The `where` condition was set earlier. Use the `setWhere()`, `andWhere()` or `orWhere()` method.');
         }
         $this->addParams($params);
         return $this;
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -717,7 +717,7 @@ class Query implements QueryInterface
         if ($this->where === null) {
             $this->where = $condition;
         } else {
-            throw new LogicException('The `where` condition was set earlier. If you want to overwrite it, use the `setWhere()` method.');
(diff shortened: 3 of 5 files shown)


lav45:
@samdark хотел показать пример реализацию защиты от ошибочного переопределения `where` критериев. Если понравится тогда продолжаю дорабатывать.

samdark:
I like and dislike it at the same time.

1. I like it because that's indeed quite common that `where()->where()` is used, and it doesn't have expected outcome to the end user. I've seen that in Yii2 projects multiple times.
2. I dislike it because there are many similar "dangerous" methods, and adding such safeguards everywhere is over-protection.
3. I dislike it because adding just a single safeguard is, as well, an inconsistency.
4. I dislike it because it doesn't read well...

Is there a way to solve the issue without such safeguard?

Tigrov:
- https://github.com/yiisoft/db/issues/714

Make `where()` an alias to `andWhere()` and add a `setWhere()` method

lav45:
@samdark check please.

src/Query/Query.php
```diff
@@ -668,7 +668,21 @@ public function union(QueryInterface|string $sql, bool $all = false): static
...
 
+    /**
+     * @throws InvalidArgumentException
+     */
     public function where(array|string|ExpressionInterface|null $condition, array $params = []): static
+    {
+        if ($this->where === null) {
+            $this->where = $condition;
+        } else {
+            throw new InvalidArgumentException('The `where` condition was set earlier. If you want to overwrite it, use the `setWhere()` method.');
```
samdark, line 679: Arguments are correct. I'd use `LogicException` or `RuntimeException`.

src/Query/QueryPartsInterface.php
```diff
@@ -580,7 +580,7 @@ public function setUnions(array $value): static;
     public function union(QueryInterface|string $sql, bool $all = false): static;
 
     /**
-     * Sets the `WHERE` part of the query.
+     * First sets the `WHERE` part of the query.
```
samdark, line 583:
```suggestion
     * Initially sets the `WHERE` part of the query.
```

src/Query/QueryPartsInterface.php
```diff
@@ -676,6 +676,18 @@ public function union(QueryInterface|string $sql, bool $all = false): static;
      */
     public function where(array|string|ExpressionInterface|null $condition, array $params = []): static;
 
+    /**
+     * Force sets the `WHERE` part of the query.
```
samdark, line 680:
```suggestion
     * Overwrites the `WHERE` part of the query.
```

samdark (changes requested):

samdark:
I like the idea :+1:

lav45:
https://github.com/yiisoft/db-pgsql/pull/401 - this PR fixed this PR)

samdark (commented):
Looks :+1: to me. Please add a line for CHANGELOG. Thanks.

samdark (approved):

Tigrov (changes requested):
Needs to add the new method to UPGRADE
https://github.com/yiisoft/db/blob/master/UPGRADE.md?plain=1#L117-L133

src/Query/Query.php
```diff
@@ -708,7 +709,21 @@ public function union(QueryInterface|string $sql, bool $all = false): static
...
 
+    /**
+     * @throws LogicException
+     */
     public function where(array|string|ExpressionInterface|null $condition, array $params = []): static
+    {
+        if ($this->where === null) {
+            $this->where = $condition;
+        } else {
+            throw new LogicException('The `where` condition was set earlier. If you want to overwrite it, use the `setWhere()` method.');
```
vjik, line 720: Mention `andWhere()` and `orWhere()` methods in exception message

src/Query/Query.php
```diff
@@ -708,7 +709,21 @@ public function union(QueryInterface|string $sql, bool $all = false): static
         return $this;
     }
 
+    /**
+     * @throws LogicException
```
vjik, line 713: Add description of exception and move this phpdoc to `QueryPartsInterface`. I think it should be part of interface.

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@
 - New #939: Add `caseSensitive` option to like condition (@vjik)
 - New #942: Allow PHP backed enums as values (@Tigrov)
 - Enh #943: Add `getCacheKey()` and `getCacheTag()` methods to `AbstractPdoSchema` class (@Tigrov)
+- Enh #944: Added `setWhere()` as method a forced for overwriting `where()` (@lav45)
+- Enh #925: Add callback to `Query::all()` and `Query::one()` methods (@Tigrov)
```
Tigrov, line 78:
```suggestion
```

UPGRADE.md
```diff
@@ -120,6 +120,7 @@ Each table column has its own class in the `Yiisoft\Db\Schema\Column` namespace
 - `QueryInterface::resultCallback()` - allows to use a callback, to be called on all rows of the query result;
 - `QueryInterface::getResultCallback()` - returns the callback to be called on all rows of the query result or 
   `null` if the callback is not set;
+- `QueryPartsInterface::setWhere()` - Overwrites the `WHERE` part of the query;
```
vjik, line 123:
```suggestion
- `QueryPartsInterface::setWhere()` - overwrites the `WHERE` part of the query;
```

src/Query/Query.php
```diff
@@ -693,7 +694,21 @@ public function union(QueryInterface|string $sql, bool $all = false): static
...
+     */
```
vjik, line 699:
```suggestion
```

src/Query/Query.php
```diff
@@ -693,7 +694,21 @@ public function union(QueryInterface|string $sql, bool $all = false): static
...
     public function where(array|string|ExpressionInterface|null $condition, array $params = []): static
+    {
+        if ($this->where === null) {
+            $this->where = $condition;
+        } else {
+            throw new LogicException('The `where` condition was set earlier. If you want to overwrite it, use the `setWhere()`, `andWhere()` or `orWhere()` method.');
```
vjik, line 705:
```suggestion
            throw new LogicException('The `where` condition was set earlier. Use the `setWhere()`, `andWhere()` or `orWhere()` method.');
```

src/Query/QueryPartsInterface.php
```diff
@@ -671,11 +672,25 @@ public function union(QueryInterface|string $sql, bool $all = false): static;
      *
      * @psalm-param ParamsType $params
      *
+     * @throws LogicException - if `where` was set previously
```
vjik, line 675:
```suggestion
     * @throws LogicException If `where` was set previously
```

vjik (approved):

vjik (approved):

src/Query/Query.php
```diff
@@ -694,6 +695,17 @@ public function union(QueryInterface|string $sql, bool $all = false): static
...
+        if ($this->where === null) {
+            $this->where = $condition;
+        } else {
+            throw new LogicException('The `where` condition was set earlier. Use the `setWhere()`, `andWhere()` or `orWhere()` method.');
+        }
+        $this->addParams($params);
+        return $this;
+    }
+
+    public function setWhere(array|string|ExpressionInterface|null $condition, array $params = []): static
```
xepozz, line 708: I'd suggest to have `clearWhere` / `dropWhere` to keep assigning logic in `where` method only. 
4  `where` methods seems confusing for me

xepozz (changes requested):

Tigrov (approved):

src/Query/Query.php
xepozz: @Tigrov nope, you should compare these:
```
$query->where($condition)->all();
$query->setWhere($condition)->all();
```

`reset` method is used only for cases when you want to clear without further overriding:

```
$query->resetWhere()->all();
$query->setWhere(null)->all();
```

in this case `reset` looks better, isn't it?

Tigrov: This is just a special case. We can make the default value in `setWhere()` equals to `null`

samdark (approved):

Tigrov:
Thanks 👍

PR #949, opened 2025-03-29
Add option for typecasting when insert or update values.

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |


samdark (approved):

samdark (approved):

src/QueryBuilder/AbstractDMLQueryBuilder.php
```diff
@@ -145,6 +147,13 @@ public function upsert(
         throw new NotSupportedException(__METHOD__ . ' is not supported by this DBMS.');
     }
 
+    public function withTypecasting(bool $typecasting = true): static
```
vjik, line 150: In the future, the possibility to get values on `INSERT`/`UPDATE` command may be added. In this case we will need PHP typecasting. So I suggest use `withDbTypecasting()` naming here and similar places.

Tigrov: No, it is not possible to get values from DB in `QueryBuilder`, it is only build a query. The query is executed inside the `Command`.

vjik: Will type casting `db` → `php` for returning from `INSERT`/`UPDATE` command values  be in `Command` class, not in `QueryBuilder`?

Tigrov: Yes, like in `Command::insertWithReturningPks()`

https://github.com/yiisoft/db/blob/7da841e7e9aa2643eaa0d0354a9d438f09b7837a/src/Command/AbstractCommand.php#L403-L408

samdark (approved):

PR #957, opened 2025-04-09
Fix Ubuntu version for MSSQL 2017

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

See https://github.com/yiisoft/db/actions/runs/14341905889/job/40202953087?pr=944

image

https://github.com/microsoft/mssql-docker/issues/924

### DIFF
--- a/.github/workflows/db-mssql.yml
+++ b/.github/workflows/db-mssql.yml
@@ -50,7 +50,7 @@ jobs:
           - php: 8.4
             mssql:
               server: 2017-latest
-              os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


Ilhomjon-coder:
may be we need to use other os runners rather than Ubuntu

we need os that supports both github and mssql 2017

samdark:
Which OS is that?

Tigrov:
Fixed https://github.com/yiisoft/db/pull/969

PR #967, opened 2025-05-01
Add `FOR` clause for query

### LINKED ISSUE #813: Add ability to create `SELECT` with `FOR UPDATE`
For example, in Cycle it implemented so: https://github.com/cycle/database/blob/331a16879523ee4209d9090c5bf20d0e62e72159/src/Query/SelectQuery.php#L119

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
Fix #813 

Related PRs:
- https://github.com/yiisoft/db-sqlite/pull/354

### DIFF
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -101,9 +101,9 @@ class Query implements QueryInterface
     protected array $with = [];
 
     /**
-     * @psalm-var list<string>|null
+     * @psalm-var list<string>
      */
-    protected ?array $for = null;
+    protected array $for = [];
 
     private bool $emulateExecution = false;
     private bool $typecasting = false;
@@ -392,7 +392,7 @@ class Query implements QueryInterface
 
     public function for(string|array|null $value): static
     {
-        if ($this->for !== null) {
+        if ($this->for !== []) {
             throw new LogicException('The `FOR` part was set earlier. Use the `setFor()` or `addFor()` method.');
         }
         return $this->setFor($value);
@@ -405,18 +405,18 @@ class Query implements QueryInterface
         }
 
         if (is_array($value)) {
-            $this->for = [...($this->for ?? []), ...$value];
+            $this->for = [...$this->for, ...$value];
             return $this;
         }
 
-        $this->for = [...($this->for ?? []), $value];
+        $this->for = [...$this->for, $value];
         return $this;
     }
 
     public function setFor(array|string|null $value): static
     {
         if ($value === null) {
-            $this->for = null;
+            $this->for = [];
             return $this;
         }
 
@@ -446,7 +446,7 @@ class Query implements QueryInterface
 
     public function getFor(): array
     {
-        return $this->for ?? [];
+        return $this->for;
     }
 
     public function getFrom(): array
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -101,9 +101,9 @@ class Query implements QueryInterface
     protected array $with = [];
 
     /**
-     * @psalm-var list<string>
+     * @psalm-var list<string>|null
      */
-    protected array $for = [];
+    protected ?array $for = null;
 
     private bool $emulateExecution = false;
     private bool $typecasting = false;
@@ -390,10 +390,33 @@ class Query implements QueryInterface
         return $this;
     }
 
-    public function for(ExpressionInterface|string|array|null $value): static
+    public function for(string|array|null $value): static
+    {
+        if ($this->for !== null) {
+            throw new LogicException('The `FOR` part was set earlier. Use the `setFor()` or `addFor()` method.');
+        }
+        return $this->setFor($value);
+    }
+
+    public function addFor(string|array|null $value): static
     {
         if ($value === null) {
-            $this->for = [];
+            return $this;
+        }
+
+        if (is_array($value)) {
+            $this->for = [...($this->for ?? []), ...$value];
+            return $this;
+        }
+
+        $this->for = [...($this->for ?? []), $value];
+        return $this;
+    }
+
+    public function setFor(array|string|null $value): static
+    {
+        if ($value === null) {
+            $this->for = null;
             return $this;
         }
 
@@ -428,7 +451,7 @@ class Query implements QueryInterface
 
     public function getFor(): array
     {
-        return $this->for;
+        return $this->for ?? [];
     }
 
     public function getFrom(): array
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -409,7 +409,7 @@ class Query implements QueryInterface
             return $this;
         }
 
-        $this->for = [...$this->for, $value];
+        $this->for[] = $value;
         return $this;
     }
 
--- a/src/Query/Query.php
+++ b/src/Query/Query.php
@@ -102,7 +102,7 @@ class Query implements QueryInterface
     protected array $with = [];
 
     /**
-     * @psalm-var list<ExpressionInterface|string>
+     * @psalm-var list<string>
      */
     protected array $for = [];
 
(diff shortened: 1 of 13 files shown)


src/Query/Query.php
```diff
@@ -385,6 +387,19 @@ public function filterWhere(array $condition): static
         return $this;
     }
 
+    public function for(ExpressionInterface|string|null $value = null): static
```
Tigrov, line 390:
```suggestion
    public function for(string|null $value = null): static
```

I think no need `ExpressionInterface` here. It should be `string` or `null` and possible constants should be provided,
for example `SelectFor::UPDATE`

src/Query/Query.php
```diff
@@ -385,6 +387,19 @@ public function filterWhere(array $condition): static
...
+    {
+        $this->for = $value;
+        return $this;
+    }
+
+    public function forUpdate(bool $value = true): static
+    {
+        return $this->for(
+            $value ? new Expression('UPDATE') : null
+        );
```
Tigrov, line 400:
```suggestion
    public function forUpdate(): static
    {
        return $this->for('UPDATE');
```
`bool $value = true` seems redundant parameter.

Use `forUpdate()` for setting `FOR UPDATE` clause and `for()` to reset it.

Tigrov (commented):
In MSSQL it is deprecated
https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/processing-select-for-update-statements?view=sql-server-ver16

MSSQL has different lock mechanism
`SELECT * FROM table_name WITH (UPDLOCK) WHERE ...`
https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver16

We can still try `SELECT FOR UPDATE` while it works

vjik:
> In MSSQL it is deprecated https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/processing-select-for-update-statements?view=sql-server-ver16
> MSSQL has different lock mechanism `SELECT * FROM table_name WITH (UPDLOCK) WHERE ...` https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver16
> We can still try `SELECT FOR UPDATE` while it works

Seems, this is about ODBC Cursor Library only: https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/appendix-f-odbc-cursor-library?view=sql-server-ver16

MSSQL supports `FOR` clause, see:
- https://learn.microsoft.com/en-us/sql/t-sql/queries/select-transact-sql?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/t-sql/queries/select-for-clause-transact-sql?view=sql-server-ver16

src/Query/Query.php
vjik: `FOR` clause may be difficult.

For example, [PostgreSQL](https://postgrespro.ru/docs/postgresql/17/sql-select):
```
SELECT ... 
[ 
	FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE } 
	[ OF _`from_reference`_ [, ...] ]
	[ NOWAIT | SKIP LOCKED ] 
	[...]
]
```

… or [MySQL](https://dev.mysql.com/doc/refman/8.4/en/select.html):

```
SELECT ... 
[
	FOR {UPDATE | SHARE}
	[OF _tbl_name_ [, _tbl_name_] ...] 
	[NOWAIT | SKIP LOCKED] | LOCK IN SHARE MODE
]
```

So, I think expression is expedient here.

src/Query/Query.php
vjik: I assumed this usage:

```php
$userConditionForUseUpdate = ...;
$query
    ->from(...)
    ->where(...)
    ->andWhere(...)
    ->forUpdate($userConditionForUseUpdate);
```

Do you think it's not need?

Tigrov:
There are only
- FOR BROWSE
- FOR XML
- FOR JSON

But no `FOR UPDATE`

src/Query/Query.php
Tigrov: The `string` is the `expression`. I do not see reasons to use `ExpressionInterface`.

It looks like index type or index method from `CommandInterface::createIndex()`

https://github.com/yiisoft/db/blob/08fe349c30ba710da793fc359bb89cc1fb87db88/src/Command/CommandInterface.php#L310-L316

src/Query/Query.php
Tigrov: Could it be

```php
$userForUpdate = ... ? 'UPDATE' : null;
$query
    ->from(...)
    ->where(...)
    ->andWhere(...)
    ->for($userForUpdate);
```

vjik: Agree. I'll fix it.

src/Query/Query.php
vjik: In future we can create special expression objects, that simplify building difficult DBMS-specific `FOR` clause with automatic table/column names quoting. WDYT?

src/Query/Query.php
vjik: done

src/Query/Query.php
Tigrov: > In future we can create special expression objects, that simplify building difficult DBMS-specific `FOR` clause with automatic table/column names quoting. WDYT?

Not sure, do you have an example?
 
> ```sql
> SELECT * FROM t1, t2 FOR SHARE OF t1 FOR UPDATE OF t2;
> ```

This can be done by
`$query->for('SHARE OF t1 FOR UPDATE OF t2');`

Tigrov:
> May be add "not supported" exception in `forUpdate()` to mssql?

Yes, for now can be marked as "not supported".

vjik:
> > May be add "not supported" exception in `forUpdate()` to mssql?
> 
> Yes, for now can be marked as "not supported".

`forUpdate()` method is in `Query` class. So we can't check it in MSSQL driver code. Maybe keep it as is?

src/Query/Query.php
vjik: OK, I removed expression support.

But array of values is keeps.  "FOR UPDATE ..." is another "for" part

```php
$query->for('SHARE OF {{t1}} FOR UPDATE OF {{t2}}');

// seems, it's more logically
$query->for([
  'SHARE OF {{t1}}',
  'UPDATE OF {{t2}}',
]);
```

src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -178,6 +179,16 @@ public function buildExpression(ExpressionInterface $expression, array &$params
...
     }
 
+    public function buildFor(array $values): string
+    {
+        $parts = array_map(
+            static fn(string $value) => 'FOR ' . $value,
+            $values,
+        );
+
+        return implode($this->separator, $parts);
```
Tigrov, line 189:
```suggestion
        if (empty($values)) {
            return '';
        }

        return 'FOR ' . implode(' FOR ', $values);
```

rustamwin (approved):
👍

src/Query/Query.php
```diff
@@ -405,6 +449,11 @@ public function getDistinct(): bool|null
         return $this->distinct;
     }
 
+    public function getFor(): array
+    {
+        return $this->for ?? [];
```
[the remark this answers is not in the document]
Tigrov: 1. I meant about property
2. Ok

Tigrov (approved):

src/Query/Query.php
```diff
@@ -385,6 +390,40 @@ public function filterWhere(array $condition): static
...
+        if ($value === null) {
+            return $this;
+        }
+
+        if (is_array($value)) {
+            $this->for = [...$this->for, ...$value];
+            return $this;
+        }
+
+        $this->for = [...$this->for, $value];
```
Tigrov, line 412:
```suggestion
        $this->for[] = $value;
```

Tigrov (approved):

Issue #725, opened 2023-07-09
Supports PHPs DateTime instances


Tigrov:
Doctrine and Cycle support PHPs `DateTime` instances

https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/cookbook/working-with-datetime.html

https://cycle-orm.dev/docs/advanced-datetime/current

It would be great if `yiisoft/db` could support PHPs `DateTime` instances for types `timestamp`, `datetime`, `date` and `time`.

```php
$command->insert('datetime_table', ['datetime' => new DateTime()])->execute();
$datetime = (new Query($db))->select(['datetime'])->from('datetime_table')->scalar();
$datetimePhpType = $tableSchema->getColumn('datetime')->phpTypecast($datetime);
$datetimePhpType instanceof DateTimeInterface;
```

I can realize if you agree.

samdark:
Yes, that is very convenient.

Tigrov:
Good, I implement

PR #968, opened 2025-05-04
Add `DateTimeColumn`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #725


src/Schema/Column/DateTimeColumn.php
```diff
@@ -0,0 +1,221 @@
...
+ * - `ColumnType::INTEGER`
+ * - `ColumnType::BIGINT`
+ * - `ColumnType::FLOAT`
+ *
+ * Possible issues:
+ * - MySQL DBMS converts `TIMESTAMP` column type values from database session time zone to UTC for storage,
+ *   and back from UTC to the session time zone when retrieve the values;
+ * - Oracle DBMS converts `TIMESTAMP WITH LOCAL TIME ZONE` column type values from database session time zone to
+ *   the database time zone for storage, and back from the database time zone to the session time zone when retrieve
+ *   the values;
```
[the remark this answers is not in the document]
Tigrov: - MySQL only has `TIMESTAMP` and `DATETIME`, both do not store time zone, but `TIMESTAMP` type has the feature mentioned above;
- Oracle also has `TIMESTAMP` and `TIMESTAMP WITH TIME ZONE`, no problems with them, but `TIMESTAMP WITH LOCAL TIME ZONE` also has the described feature.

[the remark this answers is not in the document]
Tigrov: The recommendation, set `UTC` time zone in PHP and DBMS then there will be no problems with time zones. But if you need to change the time zone, then you need to take these features into account, perhaps avoid using these types.

src/Schema/Column/DateTimeColumn.php
```diff
@@ -0,0 +1,226 @@
...
+     * The PHP time zone for the `string` datetime values when converting them to `DateTimeImmutable` objects before
+     * inserting them into the database or after retrieving them from the database. Use empty string to use current PHP
+     * time zone.
+     */
+    protected string $phpTimezone = 'UTC';
+
+    /**
+     * @psalm-suppress PropertyNotSetInConstructor
+     */
+    protected bool $shouldConvertTimezone;
```
vjik, line 71:
```suggestion
    protected ?bool $shouldConvertTimezone = null;
```

src/Schema/Column/DateTimeColumn.php
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * @psalm-var non-empty-string
+     * @psalm-suppress PropertyNotSetInConstructor
+     */
+    protected string $format;
```
vjik, line 77:
```suggestion
    /**
     * @psalm-var non-empty-string|null
     */
    protected ?string $format = null;
```

src/Schema/Column/DateTimeColumn.php
```diff
@@ -0,0 +1,226 @@
...
+    public function getPhpType(): string
+    {
+        return DateTimeImmutable::class;
+    }
+
+    /**
+     * Converts the value from the database format to PHP `DateTimeImmutable` object.
+     * If the database type does not have time zone information, the time zone will be set to the current PHP time zone.
+     *
+     * @param string|null $value
```
vjik, line 127: How can we guarantee that there will always be string or null?

Tigrov: The method is for converting values retrieved from DB and it is `string` or `null`

vjik: > The method is for converting values retrieved from DB and it is `string` or `null`

Should we change `ColumnInterface::phpTypecast()` signature to `phpTypecast(?string $value): mixed`?

Tigrov: No, for `bool`, `int` and `binary` colums DB returns `bool`, `int` and `resources`

src/Schema/Column/DateTimeColumn.php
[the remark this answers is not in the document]
samdark: For majority of cases — yes.

vjik, line 40:
```suggestion
 *   the values.
```

vjik (approved):

vjik:
It needs to add info to changelog and upgrade.

PR #974, opened 2025-05-18
Improve `dbTypecast()` method

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  |


src/Schema/Column/IntegerColumn.php
```diff
@@ -20,17 +23,21 @@ class IntegerColumn extends AbstractColumn
...
+        /** @var ExpressionInterface|int|null */
+        return match (gettype($value)) {
+            GettypeResult::INTEGER => $value,
+            GettypeResult::NULL => null,
+            GettypeResult::STRING => $value === '' ? null : (int) $value,
+            GettypeResult::DOUBLE => (int) $value,
+            GettypeResult::BOOLEAN => $value ? 1 : 0,
+            GettypeResult::OBJECT => match (true) {
                 $value instanceof ExpressionInterface => $value,
                 $value instanceof BackedEnum => (int) $value->value,
```
vjik, line 35: What if `$value->value` is empty string?

src/Schema/Column/IntegerColumn.php
```diff
@@ -20,17 +23,21 @@ class IntegerColumn extends AbstractColumn
...
-                default => (int) $value,
+                $value instanceof DateTimeInterface => $value->getTimestamp(),
+                $value instanceof Stringable => (int)(string) $value,
```
vjik, line 37: What if stringable value is empty string?

vjik (changes requested):

src/Schema/Column/BigIntColumn.php
```diff
@@ -55,4 +61,13 @@ public function phpTypecast(mixed $value): string|null
...
         return (string) $value;
     }
+
+    protected function dbTypecastString(string $value): int|string|null
+    {
+        if ($value === '') {
+            return null;
+        }
+
+        return PHP_INT_MAX >= $value && $value >= PHP_INT_MIN ? (int) $value : $value;
```
samdark, line 71: That was always quite inconvenient/dangerous part. You write the code for `int` and then for some numbers it returns `string` and everything breaks. I know that return type suggests that a string could be there but still that was the case for Yii2.

samdark: Not sure if there's anything we can do about it though. So not a blocker to merge this PR.

src/Schema/Column/DoubleColumn.php
```diff
@@ -19,13 +23,21 @@ class DoubleColumn extends AbstractColumn
...
-            return $value;
-        }
-
-        return match ($value) {
-            null, '' => null,
-            default => $value instanceof ExpressionInterface ? $value : (float) $value,
+        /** @var ExpressionInterface|float|null */
+        return match (gettype($value)) {
+            GettypeResult::DOUBLE => $value,
+            GettypeResult::INTEGER => (float) $value,
```
samdark, line 29: Should it be casted to int?

src/Schema/Column/BigIntColumn.php
Tigrov: This is typecasting for DB, some queries are faster if pass `int` instead of `string` where `int` is expected.
When typecasting for PHP it is always `string` or `null`

src/Schema/Column/DoubleColumn.php
Tigrov: I think we should return `new Param($value, DataType::STRING)`, to avoid double type detection. But this is for another PR.

samdark (approved):

vjik (approved):

Issue #748, opened 2023-08-20
Make array type only for `Constraint::$columnNames`


Tigrov:
https://github.com/yiisoft/db/blob/c5aa0b150aa2ba176a4660049ff6d98a2a92f067/src/Constraint/Constraint.php#L18

Replace with 
```php
private array $columnNames = [];
```

Related with #738

Issue #738, opened 2023-08-08
Repalce type `object` for constraint name to `Stringable`


vjik:
`object|string|null` → `Stringable|string|null`

samdark:
The reason it is like this now is that we've started DB for PHP 7.4 and `\Stringable` is available only since 8.0.

Tigrov:
Why should it accept `Stringable` objects?
Just use it as `Constraint::name((string) $value)` and accept `string|null $value` only

vjik:
Agree with @Tigrov. Seems, in all cases to `name()` pass string value.

Moreover. Seems, not need pass null to `name()`, because this method use for configure constraint once, right?

Tigrov:
For internal initialization `string` is enough.

Some class instances are initialized only in the package. Mark that methods as `@internal`?

PR #981, opened 2025-06-05
Refactor constraints

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #738, #748


src/Schema/SchemaInterface.php
```diff
@@ -390,9 +378,7 @@ public function getTableSchema(string $name, bool $refresh = false): TableSchema
      * @param bool $refresh Whether to fetch the latest available table schemas. If this is `false`, cached data may be
      * returned if available.
      *
-     * @return array The metadata for all tables in the database.
-     *
-     * @psalm-return list<TableSchemaInterface>
```
vjik, line 395: Why remove `@psalm-return list<TableSchemaInterface>`? By code of `AbstractSchema` this method always returns list.

src/Schema/AbstractSchema.php
```diff
@@ -428,21 +368,16 @@ protected function findTableNames(string $schema): array
      * @param bool $refresh Whether to fetch the latest available table metadata. If this is `false`, cached data may be
      * returned if available.
      *
-     * @throws InvalidArgumentException
-     * @throws NotSupportedException
-     *
      * @return array The metadata of the given type for all tables in the given schema.
-     *
-     * @psalm-return list<Constraint|TableSchemaInterface|array>
```
vjik, line 436: Why remove `list`? By code this method always returns `list`.

src/Schema/AbstractSchema.php
```diff
@@ -466,32 +401,34 @@ protected function getSchemaMetadata(string $schema, string $type, bool $refresh
...
         }
 
         if ($refresh || !isset($this->tableMetadata[$rawName][$type])) {
-            /** @psalm-suppress MixedArrayAssignment */
             $this->tableMetadata[$rawName][$type] = $this->loadTableTypeMetadata($type, $rawName);
             $this->saveTableMetadataToCache($rawName);
         }
 
-        /** @psalm-suppress MixedArrayAccess */
+        /** @var AbstractConstraint[]|IndexConstraint|TableSchemaInterface|null */
```
vjik: Specify it on property "tableMetaData"

src/Constraint/AbstractConstraint.php
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Represents an abstract constraint in a database.
+ *
+ * A constraint is a rule that's enforced on the data in a table, such as a primary key, a foreign key, a default value,
+ * check or index constraint.
+ *
+ * It's the base class for all constraint classes, and defines methods that are common and shared by all constrained
+ * classes, with name and columnNames being the most common ones.
+ */
+abstract class AbstractConstraint
```
vjik, line 16: Are setters needed in these classes?

May be keep only constructors with public read only properties?

vjik (changes requested):

src/Schema/SchemaInterface.php
Tigrov: To avoid a lot of annotations.
The annotation `@return TableSchemaInterface[]` tells enough to understand what type it is and also enough for psalm to recognize it.

src/Schema/AbstractSchema.php
Tigrov: Changed to `@return AbstractConstraint[][]|IndexConstraint[]|TableSchemaInterface[]`

src/Constraint/AbstractConstraint.php
Tigrov: I think there should be an option to change the values. This may be required by other issues and future opportunities (mostly in `db-migration`).

vjik (approved):

PR #984, opened 2025-06-09
Add `ConnectionInterface::select()` method

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,7 @@
 - Chg #980: Add constructor with DB connection to `AbstractCommand` (@vjik)
 - Enh #979: Allow `ExpressionInterface` for column definitions when create table (@Tigrov)
 - Enh #982: Reduce binding parameters (@Tigrov)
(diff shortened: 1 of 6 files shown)


samdark:
Codewise it's OK but interface-wise I'm not 100% sure it belongs to connection. Rather not...

vjik:
> Codewise it's OK but interface-wise I'm not 100% sure it belongs to connection. Rather not...

May be so:

`public function createQuery(): QueryInterface;`

Seems, this variant is more consistent with `createCommand()`.

Tigrov:
> How about insert, update etc. etc?

Yes, if it will create `InsertQuery` and `UpdateQuery` instances but for now we have only `Query` which should be named `SelectQuery` or `Select`.

vjik (approved):

PR #986, opened 2025-06-11
Replace SQLite quotes from ` to "

Tigrov:
### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/362
- https://github.com/yiisoft/active-record/pull/430

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️/❌
| Fixed issues  |

### DIFF
--- a/tests/Support/DbHelper.php
+++ b/tests/Support/DbHelper.php
@@ -98,8 +98,8 @@ final class DbHelper
     public static function replaceQuotes(string $sql, string $driverName): string
     {
         return match ($driverName) {
-            'mysql', 'sqlite' => str_replace(['[[', ']]'], '`', $sql),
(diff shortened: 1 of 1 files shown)


vjik (commented):
SQLite recommends use double-quotes for identifiers.

` also is supported, but "This is not standard SQL. This quoting mechanism is used by MySQL and is included in SQLite for compatibility."

[SQLite docs](https://www.sqlite.org/lang_keywords.html)

I suggest keep double-quotes usage.

Tigrov:
Currently `db-sqlite` users backticks ` to quote identifiers, the PR changes it to double quote " as recommended by the SQLite doc

Issue #95, opened 2016-10-19
DB CASE WHEN in SELECT


mariuszkrzaczkowski:
New feature, adding the possibility to use CASE WHEN in yii\db\Query
## DB support for CASE WHEN
- [x] MySQL
- [x] PostgreSQL
- [ ] MSSQL
- [ ] Oracle
- [x] SQLite
- [ ] Cubrid
## TODO
- [ ] Choose syntax for it
- [ ]  Write tests
- [ ] Implement

quackadillyblip:
Would it be possible to nest Query's into the THEN clause? 
for example
```
select (CASE WHEN type=1 THEN (
select quantity from exampletable where example_id=example.id
)
CASE WHEN type=2 THEN (
select quantity from othertable where example_id=example.id
)
ELSE 0)
as quantity
from example
```

Tigrov:
Done

PR #988, opened 2025-06-12
Add `CaseExpression`

Tigrov:
### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/363
- https://github.com/yiisoft/db-mysql/pull/400
- https://github.com/yiisoft/db-pgsql/pull/415
- https://github.com/yiisoft/db-mssql/pull/367
- https://github.com/yiisoft/db-oracle/pull/331

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #95


src/Expression/WhenClause.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Expression;
+
+/**
+ * Represents the condition and the result of a WHEN clause in a SQL CASE statement.
+ *
+ * @see CaseExpression
+ */
+final class WhenClause
```
vjik, line 15: May be move `CaseExpression`, `CaseExpressionBuilder` and `WhenClause` to `Yiisoft\Db\Expression\Case` namespace?

src/Expression/CaseExpression.php
```diff
@@ -0,0 +1,166 @@
...
+    {
+    }
+
+    /**
+     * Adds a condition and its corresponding result to the CASE expression.
+     *
+     * @param array|bool|ExpressionInterface|float|int|string $when The condition to check (WHEN).
+     * @param bool|ExpressionInterface|float|int|string|null $then The result to return if the condition is `true` (THEN).
+     */
+    public function addWhen(
```
vjik: Are setters needed here? May create public read only properties?

samdark: I think it's a good idea since the usage could be chained:

```php
$case = (new CaseExpression('expression'))
    ->addWhen(1, 'result1')
    ->addWhen(2, 'result2')
    ->else('defaultResult');
```

How would it look like with read only properties?

samdark (approved):

src/Expression/WhenClause.php
Tigrov: I thought to rearrange all expressions later. Perhaps move it to `Yiisoft\Db\Expression\Statement` namespace.

src/Expression/CaseExpression.php
vjik:
```php
$case = new CaseExpression(
    case: 'expression'
    when: [
        new WhenClause(1, 'result1'),
        new WhenClause(2, 'result2'),
    ],
    else: 'defaultResult',
);
```

Tigrov: It will require to prepare all operands before create the expression.

Currently it looks more flexible.

```php
$case = new CaseExpression();
foreach ($values as $value) {
    $case->addWhen("$value > 0", $value);
}
```

vjik: > It will require to prepare all operands before create the expression.

And this is good, because it guarantee, that `CaseExpression` instance will be in correct state. `CaseExpression` is simple and should build in one place.

```php
$case = new CaseExpression();
// Here `$case` is incorrect from the point of application logic.
    
foreach ($values as $value) {
    $case->addWhen("$value > 0", $value);
    // Here `$case` is incorrect from the point of application logic.
}

// Here `$case` is correct
```

I think this is better:

```php
$case = new CaseExpression(
    case: 'expression'
    when: array_map(
        static fn ($value) => new WhenClause("$value > 0", $value),
        $values,
    ),
    else: 'defaultResult',
);
```

Tigrov: Look at this code and compare it with the suggested approach.

```php
$greatest = new Greatest();
$case = (new CaseExpression($greatest))->else(array_pop($operands));

foreach ($operands as $operand) {
    $greatest->addOperand($operand);
    $case->addWhen(new Length($operand), $operand);
}
```

vjik: What is "Greatest" in this example?

In general, code looks complexity. First, `Greatest` instance is created and passed to `CaseExpresion` and only then `Greatest` was configured.

src/Expression/CaseExpression.php
```diff
@@ -0,0 +1,166 @@
...
```
rustamwin:
```suggestion
    public function when(
```
That's better, isn't it?

src/Expression/CaseExpression.php
Tigrov: `Greatest` and `Length` are other expressions not yet present in `db`

To avoid confusion, it could be like this:

```php
$greatest = new Greatest();
$case = (new CaseExpression())->else(array_pop($operands));

foreach ($operands as $operand) {
    $length = new Length($operand);
    $greatest->addOperand($length);
    $case->addWhen($length, $operand);
}

$case->case($greatest);
```

rustamwin (approved):

src/Expression/CaseExpression.php
Tigrov: At first glance it seems better, but it is not clear whether this sets a new value or adds another value.

src/Expression/CaseExpression.php
vjik: I would prefer to split creating of operands for `Greatest` and clauses for `CaseExpression`:

```php
$lengths = array_map(
    static fn ($operand) => new Length($operand),
    $operands,
),
$lastLength = array_pop($length);

new CaseExpression(
    case: new Greatest(operands: $lengths),
    whenClauses: array_map(
        static fn($length) => new WhenClause($length, $length->operand),
        $lengths,
    ),
    else: $lastLength->operand,
);
```

In fact, in this case, setters are only needed to create and configure the object, which seems like the responsibility of a constructor.

src/Expression/CaseExpression.php
```diff
@@ -0,0 +1,209 @@
...
+     * @var WhenClause[] List of pairs of conditions and their corresponding results in the CASE expression.
+     */
+    private array $whenClauses = [];
+    /**
+     * @var mixed The result to return if no conditions match in the CASE expression.
+     * If not set, the CASE expression will not have an ELSE clause.
+     *
+     * @psalm-suppress PropertyNotSetInConstructor
+     */
+    private mixed $else;
```
vjik, line 69: Move `$else` and `$whenClause` to constructor. This allow to use `CaseExpression` without call setters for cases when all the data is known in advance.

vjik (approved):
Add line to changelog

PR #1003, opened 2025-07-20
Refactor condition objects

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️


src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -504,8 +504,8 @@ protected function defaultConditionClasses(): array
     {
         return [
             'NOT' => Condition\NotCondition::class,
-            'AND' => Condition\AndCondition::class,
-            'OR' => Condition\OrCondition::class,
+            'AND' => Condition\Conjunction\AndCondition::class,
+            'OR' => Condition\Conjunction\OrCondition::class,
```
samdark: Do we need `Conjunction` as part of namespace? If we do, I think it's called `Logical`, not conjunction.

src/QueryBuilder/Condition/Conjunction/ExpressionsConjunctionBuilder.php
```diff
@@ -19,57 +17,57 @@
...
      *
      * @throws Exception
      * @throws InvalidArgumentException
      * @throws InvalidConfigException
      * @throws NotSupportedException
+     *
+     * @psalm-param array<array|ExpressionInterface|scalar> $expressions
      */
-    public function build(ExpressionInterface $expression, array &$params = []): string
+    public function build(array $expressions, array &$params = []): string
```
Tigrov: Why need this builder?

It could be `AbstractConditionBuilder` and concrete builders containing the operator
```php
final class AndConditionBuilder extends AbstractConditionBuilder
{
   private const OPERATOR = 'AND';
}
```

src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -516,8 +516,8 @@ protected function defaultConditionClasses(): array
             'OR NOT LIKE' => Condition\LikeCondition::class,
             'EXISTS' => Condition\ExistsCondition::class,
             'NOT EXISTS' => Condition\ExistsCondition::class,
-            'ARRAY OVERLAPS' => Condition\ArrayOverlapsCondition::class,
-            'JSON OVERLAPS' => Condition\JsonOverlapsCondition::class,
+            'ARRAY OVERLAPS' => Condition\Overlaps\ArrayOverlapsCondition::class,
+            'JSON OVERLAPS' => Condition\Overlaps\JsonOverlapsCondition::class,
```
Tigrov: I also think there is no need to separate two conditions in different namespaces (as for `Conjunction`).

src/QueryBuilder/Condition/HashCondition.php
```diff
@@ -9,16 +9,12 @@
...
-
     /**
-     * @return array|null The condition specification.
+     * @param array|null $hash The condition specification.
      */
-    public function getHash(): array|null
-    {
-        return $this->hash;
+    public function __construct(
+        public readonly array|null $hash = [],
```
Tigrov:
```suggestion
        public readonly array $hash = [],
```

Looks like it should be only `array`.

src/QueryBuilder/Condition/Conjunction/ExpressionsConjunctionBuilder.php
vjik: Composition is preferable to inheritance.

src/QueryBuilder/AbstractDQLQueryBuilder.php
vjik: image

I grouped classes that contains more than condition and builder.
Would it be better if everything was in the same namespace?

src/QueryBuilder/AbstractDQLQueryBuilder.php
vjik: Ok, i'll rename it.

src/QueryBuilder/Condition/HashCondition.php
vjik: Ok, i'll do it.

vjik: Done

src/QueryBuilder/Condition/Conjunction/ExpressionsConjunctionBuilder.php
Tigrov: Depends on the case.
Compare `$expression` class name with `AndCondition` and `OrCondition` to get operator. Then `AndConditionBuilder` and `OrConditionBuilder` are not needed.

src/QueryBuilder/AbstractDQLQueryBuilder.php
vjik: done

src/QueryBuilder/Condition/Conjunction/ExpressionsConjunctionBuilder.php
vjik: done

PR #1005, opened 2025-07-22
Refactor `DSN` classes

Tigrov:
### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/373
- https://github.com/yiisoft/db-mysql/pull/410
- https://github.com/yiisoft/db-pgsql/pull/427
- https://github.com/yiisoft/db-mssql/pull/378
- https://github.com/yiisoft/db-oracle/pull/344

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  |


src/Connection/AbstractDsn.php
```diff
@@ -5,42 +5,63 @@
...
- * It also allows you to access individual components of the DSN, such as the driver or the database name.
+ * For DSN string format use `(string)` type casting operator.
+ *
+ * It's typically used to build a DSN string, which has all the necessary information to connect to a database,
+ * such as the database driver, hostname, database name, port, and options.
+ *
+ * It also allows you to access individual components of the DSN via public readonly properties.
  */
-abstract class AbstractDsn implements DsnInterface, Stringable
+abstract class AbstractDsn implements Stringable
```
vjik: May be remove `AbstractDsn` also?

Tigrov: Thought about it. It is used in MySQL and Postgres. Do you think it's better to copy the code there?

vjik: I think, yes. If most DBMS have their own implementation, so abstract DSN makes little sense.

samdark (approved):
Overall, that's a good move 👍. We lose the ability to get components of the DSN directly though, but if we assume that all DSNs follow the same naming schema (is that true? 🤔) then, if needed, we can parse it.

PR #1009, opened 2025-07-30
Ability to configure columns

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

Example of configuration:
```php
ConnectionInterface::class => [
    'class' => Connection::class,
    '__construct()' => [
        'driver' => new Driver(
            $params['db']['dsn'],
            $params['db']['username'],
            $params['db']['password']
        ),
        'columnFactory' => new ColumnFactory(
            definitions: [
                ColumnType::ARRAY => ArrayLazyColumn::class,
                ColumnType::DATETIME => [
                    DateTimeColumn::class,
                    'dbTimezone' => '+02:00',
                    'phpTimezone' => 'Europe/Berlin',
                ],
            ]
        )
    ],
],
```


src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -72,10 +72,12 @@ abstract class AbstractColumnFactory implements ColumnFactoryInterface
      *
      * @psalm-param ColumnClassMap $columnClassMap
      * @psalm-param TypeMap $typeMap
+     * @psalm-param array<ColumnType::*, ColumnInfo> $defaultColumnsParams
      */
     public function __construct(
         protected array $columnClassMap = [],
         protected array $typeMap = [],
+        protected array $defaultColumnsParams = [],
```
samdark:
```suggestion
        protected array $defaults = [],
```

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -70,13 +83,33 @@ abstract class AbstractColumnFactory implements ColumnFactoryInterface
...
+                unset($value[0]);
+                if (!empty($value)) {
+                    $resultColumnClassDefaults[$type] = $value;
+                }
+                continue;
+            }
+            $resultColumnClassMap[$type] = $value;
+        }
+        $this->columnClassMap = $resultColumnClassMap;
+        $this->columnClassDefaults = $resultColumnClassDefaults;
```
Tigrov: Looks better to add `protected readonly array $columnClassDefaults` to constructor as you did in the first commit.

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -38,6 +41,16 @@ abstract class AbstractColumnFactory implements ColumnFactoryInterface
      */
     protected const TYPE_MAP = [];
 
+    /**
+     * @psalm-var ColumnClassMap
```
samdark, line 44: phpdoc needed.

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -38,6 +41,16 @@ abstract class AbstractColumnFactory implements ColumnFactoryInterface
...
+     */
+    protected readonly array $columnClassMap;
+
+    /**
+     * @psalm-var array<ColumnType::*, ColumnInfo>
```
samdark: phpdoc needed.

samdark (commented):
Need CHANGELOG and docs update as well.

samdark (approved):
👍 but need docs about it.

src/Schema/Column/AbstractColumnFactory.php
vjik: Done

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -166,13 +168,14 @@ public function fromType(string $type, array $info = []): ColumnInterface
         }
 
         /** @psalm-var class-string<ColumnInterface> $columnClass */
-        $columnClass = $this->mapType($this->columnClassMap, $type, $info)
+        $columnClass = $this->mapType($this->classMap, $type, $info)
             ?? $this->getColumnClass($type, $info);
 
-        $column = new $columnClass($type, ...$info);
+        $columnParams = $info + ($this->classDefaults[$type] ?? []);
```
Tigrov, line 210: Maybe use `$columnClass` instead of `$type` to set/get default parameters?

`$columnClass` also depends on `$info` and can be different for the same `$type`.

vjik: > `$columnClass` also depends on `$info` and can be different for the same `$type`.

That's why I used `$type`. For example, one class used for two types, and user want to set different defaults for each type. 

We set column classes by types, so class defaults also must be set by types.

Tigrov (approved):

Tigrov (approved):
👍

PR #1013, opened 2025-08-04
Change `BinaryColumn::phpTypecast()` to return `StringableSteam|string|null`

Tigrov:
### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/378
- https://github.com/yiisoft/db-pgsql/pull/432
- https://github.com/yiisoft/db-mssql/pull/384
- https://github.com/yiisoft/db-oracle/pull/349

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  |


src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @return string The result of reading the resource stream.
+     */
+    public function __toString(): string
+    {
+        /**
+         * @psalm-suppress PossiblyFalsePropertyAssignmentValue, PossiblyInvalidArgument
+         * @var string
+         */
+        return match (gettype($this->value)) {
```
vjik, line 70: Add available types of value in constructor phpdoc

src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+            GettypeResult::RESOURCE => $this->value = stream_get_contents($this->value),
+            GettypeResult::RESOURCE_CLOSED => throw new InvalidArgumentException('Resource is closed.'),
```
vjik:
```suggestion
            GettypeResult::RESOURCE_CLOSED => throw new RuntimeException('Resource is closed.'),
```

src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+            default => $this->value,
+        };
+    }
+
+    /**
+     * @return resource|string The resource stream or the result of reading the stream.
+     */
+    public function getValue(): mixed
```
vjik, line 80: Use public readonly value instead of getter.

src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+    public function __toString(): string
+    {
+        /**
+         * @psalm-suppress PossiblyFalsePropertyAssignmentValue, PossiblyInvalidArgument
+         * @var string
+         */
+        return match (gettype($this->value)) {
+            GettypeResult::RESOURCE => $this->value = stream_get_contents($this->value),
+            GettypeResult::RESOURCE_CLOSED => throw new InvalidArgumentException('Resource is closed.'),
+            default => $this->value,
```
vjik, line 73: Should we throw `LogicException` when value is not resource?

src/Schema/Data/StringableStream.php
Tigrov: It is changeable. When reading a stream, the result is stored in the `$value` property.

src/Schema/Data/StringableStream.php
Tigrov: I suppose the `resource` PHPDoc type includes open resource and closed resource.

src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+ * // @var resource $resource
+ * $stream = new StringableStream($resource);
+ *
+ * echo $stream;
+ * ```
+ */
+final class StringableStream implements Stringable
+{
+    /**
+     * @param resource|string $value The resource stream or the result of reading the stream.
```
vjik: Should class created for resource only?

src/Schema/Data/StringableStream.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function __construct(private mixed $value)
+    {
```
vjik, line 39: should we check resource type in constructor?

src/Schema/Data/StringableStream.php
Tigrov: Yes it should.

src/Schema/Data/StringableStream.php
Tigrov: I don't think it's necessary. It is used internally but if users will use the class they should use certain types described in constructor doc.

src/Schema/Data/StringableStream.php
vjik: So need to use only resource in phpdoc and replace  promoted property to classic property with resource|string type.

Tigrov: Done

vjik (approved):

PR #1014, opened 2025-08-04
Replace "escaping replacements" methods with "escape" constructor parameter in `Like`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Relative PRs:
- https://github.com/yiisoft/db-oracle/pull/350

### DIFF
--- a/src/QueryBuilder/Condition/Like.php
+++ b/src/QueryBuilder/Condition/Like.php
@@ -8,7 +8,6 @@ use Iterator;
 use InvalidArgumentException;
 use Yiisoft\Db\Expression\ExpressionInterface;
 
-use function array_key_exists;
 use function is_array;
 use function is_int;
 use function is_string;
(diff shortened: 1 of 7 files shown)


src/QueryBuilder/Condition/Like.php
```diff
@@ -18,46 +17,26 @@
...
      * behavior.
+     * @param bool $escape Whether to escape the value. Defaults to `true`. If `false`, the value will be used as is
+     * without escaping.
      */
     public function __construct(
         public readonly string|ExpressionInterface $column,
         public readonly string $operator,
         public readonly array|int|string|Iterator|ExpressionInterface|null $value,
         public readonly ?bool $caseSensitive = null,
+        public readonly bool $escape = self::DEFAULT_ESCAPE,
```
Tigrov, line 36:
```suggestion
        public readonly bool $escape = true,
```

Why not to use `true` directly?

vjik: Value is used in `fromArrayDefinition` method also.

Tigrov: 👌

PR #1015, opened 2025-08-05
Add mode parameter to `Like` condition

### LINKED ISSUE #811: Add ability to create like condition as `%value`, `value%` and `value` without `%`
Currently, is avaliable only when manually escape value. See here: https://github.com/yiisoft/db/blob/03dc4342c5e0943e21c7307d01ca708258a2a105/src/QueryBuilder/Condition/Builder/LikeConditionBuilder.php#L87

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
Fix #811

### DIFF
--- a/src/QueryBuilder/Condition/Like.php
+++ b/src/QueryBuilder/Condition/Like.php
@@ -29,7 +29,7 @@ final class Like implements ConditionInterface
      * behavior.
      * @param bool $escape Whether to escape the value. Defaults to `true`. If `false`, the value will be used as is
      * without escaping.
-     * @param LikeMode $mode The mode for the LIKE operation (contains, starts with, ends with).
+     * @param LikeMode $mode The mode for the LIKE operation (contains, starts with, ends with or custom pattern).
      */
     public function __construct(
         public readonly string|ExpressionInterface $column,
--- a/src/QueryBuilder/Condition/Like.php
+++ b/src/QueryBuilder/Condition/Like.php
@@ -11,6 +11,7 @@ use Yiisoft\Db\Expression\ExpressionInterface;
 use function is_array;
 use function is_int;
 use function is_string;
+use function sprintf;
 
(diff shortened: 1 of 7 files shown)


tests/Db/QueryBuilder/Condition/Builder/LikeBuilderTest.php
```diff
@@ -28,4 +35,23 @@ public function testOperatorException(): void
...
+
+        $params = [];
+        $likeBuilder->build($likeCondition, $params);
+
+        assertCount(1, $params);
+
+        /** @var Param $param */
+        $param = reset($params);
+        assertSame($expected, $param->value);
+        assertSame(DataType::STRING, $param->type);
```
Tigrov:
```suggestion
        $this->assertCount(1, $params);

        /** @var Param $param */
        $param = reset($params);
        $this->assertSame($expected, $param->value);
        $this->assertSame(DataType::STRING, $param->type);
```

Use one style in the tests?

vjik: done

src/QueryBuilder/Condition/Like.php
```diff
@@ -27,13 +29,15 @@ final class Like implements ConditionInterface
      * behavior.
      * @param bool $escape Whether to escape the value. Defaults to `true`. If `false`, the value will be used as is
      * without escaping.
+     * @param LikeMode $mode The mode for the LIKE operation (contains, starts with, ends with).
```
Tigrov:
```suggestion
     * @param LikeMode $mode The mode for the LIKE operation (contains, starts with, ends with or custom pattern).
```

Tigrov (approved):

PR #1017, opened 2025-08-06
Split `Between` condition to `Between` and `NotBetween`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️


src/QueryBuilder/Condition/Builder/BetweenBuilder.php
```diff
@@ -38,7 +38,7 @@ public function __construct(private readonly QueryBuilderInterface $queryBuilder
      */
     public function build(ExpressionInterface $expression, array &$params = []): string
     {
-        $operator = $expression->operator;
+        $operator = $expression->isNot() ? 'NOT BETWEEN' : 'BETWEEN';
```
Tigrov: Get operator from class name like for `AndX` and `OrX`
```php
$operator = match ($expression::class) {
    Between::class => 'BETWEEN',
    NotBetween::class => 'NOT BETWEEN',
};
```

Tigrov (approved):

src/QueryBuilder/Condition/Builder/BetweenBuilder.php
vjik: Done.

Tigrov (approved):
Missing test for `NotBetween`

src/QueryBuilder/Condition/AbstractBetween.php
```diff
@@ -0,0 +1,64 @@
...
+            $operands[2],
+        );
+    }
+
+    /**
+     * Validates the given column to be string or `ExpressionInterface`.
+     *
+     * @throws InvalidArgumentException If the column isn't a string or `ExpressionInterface`.
+     */
+    private static function validateColumn(string $operator, mixed $column): string|ExpressionInterface
```
Tigrov, line 54:
```suggestion
    private static function validateColumn(mixed $column): string|ExpressionInterface
```

vjik: Why? `$operator` is used in exception message.

Tigrov: 👌

PR #1018, opened 2025-08-06
Split `BetweenColumns` condition to `BetweenColumns` and `NotBetweenColumns`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️

### DIFF
--- a/src/QueryBuilder/Condition/AbstractBetweenColumns.php
+++ b/src/QueryBuilder/Condition/AbstractBetweenColumns.php
@@ -1,0 +1,138 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\QueryBuilder\Condition;
+
+use Iterator;
+use InvalidArgumentException;
+use Yiisoft\Db\Expression\ExpressionInterface;
+
+use function is_array;
+use function is_int;
+use function is_string;
+
+/**
+ * @internal
+ *
+ * Represents a `BETWEEN` and `NOT BETWEEN` operator where values are between two columns.
+ *
+ * For example:
+ *
+ * ```php
+ * new BetweenColumnsCondition(42, 'min_value', 'max_value')
+ * // Will be build to:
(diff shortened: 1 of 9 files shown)


src/QueryBuilder/Condition/AbstractBetweenColumns.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Validates the given value to be `array`, `int`, `string`, `Iterator` or `ExpressionInterface`.
+     *
+     * @throws InvalidArgumentException If the value isn't `array`, `int`, `string`, `Iterator` or `ExpressionInterface`.
+     */
+    private static function validateValue(
+        string $operator,
+        mixed $value
+    ): array|int|string|Iterator|ExpressionInterface {
```
Tigrov: Is it possible also be float?

src/QueryBuilder/Condition/AbstractBetweenColumns.php
```diff
@@ -0,0 +1,117 @@
...
+        if (
+            is_array($value) ||
+            is_int($value) ||
+            is_string($value) ||
+            ($value instanceof Iterator) ||
```
Tigrov:
```suggestion
            is_iterable($value) ||
            is_int($value) ||
            is_string($value) ||
```

Or improve using `match (gettype()) {...}`

src/QueryBuilder/Condition/AbstractBetweenColumns.php
vjik: I don't know. Previous logic is kept. I suggest fix it in separate PR

Tigrov: 👌

Tigrov (approved):

PR #1027, opened 2025-08-12
Add compare conditions

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


src/QueryBuilder/Condition/NotEquals.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\QueryBuilder\Condition;
+
+/**
+ * Represents a not equals condition like `"column" <> value`.
```
Tigrov, line 8: Is operator <> better than != ?

vjik: `<>` in SQL standard, but `!=` is not.

PR #1031, opened 2025-08-14
Optimize SQL generation for `Not` condition

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/QueryBuilder/Condition/Builder/NotBuilder.php
```diff
@@ -36,19 +53,66 @@ public function __construct(private readonly QueryBuilderInterface $queryBuilder
...
+            ),
+            NotLike::class => new Like(
+                $condition->column,
+                $condition->value,
+                $condition->caseSensitive,
+                $condition->escape,
+                $condition->mode,
+                $condition->conjunction,
+            ),
+            Not::class => $condition->condition,
```
Tigrov, line 114: What if `$condition->condition === null`?

Tigrov (approved):

src/QueryBuilder/Condition/Builder/NotBuilder.php
```diff
@@ -36,19 +53,66 @@ public function __construct(private readonly QueryBuilderInterface $queryBuilder
...
+            }
+        }
 
-        return "{$this->getNegationOperator()} ($expressionValue)";
+        $sql = $this->queryBuilder->buildCondition($condition, $params);
+        return "NOT ($sql)";
     }
 
-    protected function getNegationOperator(): string
+    protected function tryCreateNegatedCondition(ConditionInterface $condition): array|string|ExpressionInterface|null
```
samdark:
```suggestion
    protected function createNegatedCondition(ConditionInterface $condition): array|string|ExpressionInterface|null
```

because try is usually "do something and throw exception if failed".

src/QueryBuilder/Condition/Builder/NotBuilder.php
vjik: Good catch! I fixed it. Review it.

src/QueryBuilder/Condition/Builder/NotBuilder.php
vjik: I was thinking about `enum::tryFrom()`...

vjik: done

PR #1033, opened 2025-08-14
Add `All` and `None` conditions

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/QueryBuilder/Condition/Builder/AllBuilder.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Builds SQL expressions for {@see All} condition.
+ *
+ * @implements ExpressionBuilderInterface<All>
+ */
+final class AllBuilder implements ExpressionBuilderInterface
+{
+    public function build(ExpressionInterface $expression, array &$params = []): string
+    {
+        return '';
```
samdark, line 20: I'd use `true` in this case because it could be the only condition and in this case invalid SQL will be created for empty string:

```sql
SELECT *
FROM X
WHERE true
```

vs 

```sql
SELECT *
FROM X
WHERE
```

vjik: This case should correct handle. Some conditions sometimes returns empty string.

Tigrov: Actually it should be
```sql
SELECT *
FROM X
```

vjik: That's how it works. See https://github.com/yiisoft/db/pull/1033/commits/931802d66954c6645d8e1cd779fe9dbf4abe42ea

vjik:
> In what cases should this be used?

Any places where we create conditions. For example, when result is "no filter" we can return `All`.

PR #1043, opened 2025-08-24
Update `DateTimeValue`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Expression/Value/Builder/DateTimeValueBuilder.php
+++ b/src/Expression/Value/Builder/DateTimeValueBuilder.php
@@ -4,21 +4,12 @@ declare(strict_types=1);
 
 namespace Yiisoft\Db\Expression\Value\Builder;
 
-use DateTimeImmutable;
-use DateTimeInterface;
-use InvalidArgumentException;
-use Stringable;
 use Yiisoft\Db\Constant\ColumnType;
-use Yiisoft\Db\Constant\GettypeResult;
 use Yiisoft\Db\Expression\Builder\ExpressionBuilderInterface;
 use Yiisoft\Db\Expression\Value\DateTimeValue;
 use Yiisoft\Db\Expression\ExpressionInterface;
 use Yiisoft\Db\QueryBuilder\QueryBuilderInterface;
 use Yiisoft\Db\Schema\Column\ColumnFactoryInterface;
-
-use function date_create_immutable;
-use function gettype;
-use function sprintf;
 
 /**
  * Builder for {@see DateTimeValue} expressions.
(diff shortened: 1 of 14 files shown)


src/Expression/Value/Builder/DateTimeValueBuilder.php
```diff
@@ -43,41 +34,23 @@ public function __construct(
...
-            (string) (int) $value => DateTimeImmutable::createFromFormat('U', $value),
-            (string) (float) $value => DateTimeImmutable::createFromFormat('U.u', $value),
-            default => date_create_immutable($value),
+        return match ($type) {
+            ColumnType::TIMESTAMP,
+            ColumnType::TIME,
+            ColumnType::TIMETZ,
+            ColumnType::DATETIME,
+            ColumnType::DATETIMETZ => $type,
+            default => ColumnType::TIMESTAMP,
```
vjik, line 53: I think we should restrict type values in `DateTimeValue` and don't use Timestamp column as default.

Tigrov: 1) Good idea to restrict type values in `DateTimeValue`
2) Which type use by default?

vjik: Here default throw exception, in `DateTimeValue` use `DATETIMETZ` as default.

Tigrov: Currently possible do `new DateTimeValue('2025-08-24 10:30:00', ColumnType::INTEGER)` to save value into integer column. If limit possible types it will be impossible.

vjik: > Currently possible do `new DateTimeValue('2025-08-24 10:30:00', ColumnType::INTEGER)` to save value into integer column. If limit possible types it will be impossible.

In this case user should pass type by `$info`:

```php
new DateTimeValue('2025-08-24 10:30:00', ColumnType::TIMESTAMP, ['type' => ColumnType::INTEGER])
```

This approach is consistent with working with column classes in Yii DB.

Tigrov: Looks complicated, maybe swap types and create column from the type inside `$info`, default `ColumnType::TIMESTAMP`

```php
new DateTimeValue('2025-08-24 10:30:00', ColumnType::INTEGER, ['type' => ColumnType::TIMESTAMP])
```
the same
```php
new DateTimeValue('2025-08-24 10:30:00', ColumnType::INTEGER)
```

PR #1053, opened 2025-09-01
Refactor `CaseExpression`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | 

### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/392
- https://github.com/yiisoft/db-mysql/pull/429
- https://github.com/yiisoft/db-pgsql/pull/448
- https://github.com/yiisoft/db-mssql/pull/399
- https://github.com/yiisoft/db-oracle/pull/366


src/Expression/Statement/CaseX.php
```diff
@@ -0,0 +1,120 @@
...
+     * If not provided, the `CASE` expression will be a WHEN-THEN structure without a specific case value.
+     * @param ColumnInterface|string $valueType Optional data type of the CASE expression which can be used in some DBMS
+     * to specify the expected type (for example in PostgreSQL).
+     * @param mixed|When ...$args List of `WHEN` conditions and their corresponding results represented
+     * as {@see When} instances or `ELSE` value in the `CASE` expression.
+     */
+    public function __construct(
+        public readonly mixed $value = null,
+        public readonly string|ColumnInterface $valueType = '',
+        mixed ...$args,
```
vjik, line 90: May be explicitly declare parameters `When[] $when`  and `mixed $else` ?

Tigrov: `$else` can be omitted if no else. And `null` is not suitable because the value can be `null`.

PR #1063, opened 2025-09-16
Fix batch result keys and callback

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

Related PRs:

- https://github.com/yiisoft/active-record/pull/454

### DIFF
--- a/src/Query/BatchQueryResult.php
+++ b/src/Query/BatchQueryResult.php
@@ -89,15 +89,16 @@ final class BatchQueryResult implements BatchQueryResultInterface
 
         $this->dataReader ??= $this->query->createCommand()->query()->indexBy($this->indexBy);
 
+        $isContinuousIndex = $this->indexBy === null;
+        $startIndex = $isContinuousIndex ? ($this->index + 2) * $this->batchSize : 0;
+
         for (
-            $count = 0;
-            $count < $this->batchSize && $this->dataReader->valid();
-            ++$count, $this->dataReader->next()
+            $leftCount = $this->batchSize;
+            $leftCount > 0 && $this->dataReader->valid();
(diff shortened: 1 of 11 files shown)


src/Query/BatchQueryResult.php
```diff
@@ -74,21 +81,25 @@ private function fetchData(): array
...
 
-        $this->dataReader ??= $this->query->createCommand()->query();
+        $this->dataReader ??= $this->query->createCommand()->query()->indexBy($this->indexBy);
 
         for (
-            $leftCount = $this->batchSize;
-            $leftCount > 0 && $this->dataReader->valid();
-            --$leftCount, $this->dataReader->next()
+            $count = 0;
+            $count < $this->batchSize && $this->dataReader->valid();
```
Tigrov: Comparison with a constant was faster.

src/Query/BatchQueryResult.php
```diff
@@ -74,21 +81,25 @@ private function fetchData(): array
...
+            ++$count, $this->dataReader->next()
         ) {
+            /** @var int|string $key */
+            $key = $this->indexBy === null
+                ? ($this->index + 1) * $this->batchSize + $count
```
Tigrov: Should be taken out of the cycle

src/Query/BatchQueryResult.php
vjik: Refactored.

src/Query/BatchQueryResult.php
vjik: Refactored.

Tigrov (approved):

PR #1038, opened 2025-08-18
Add ability to pass `FROM` clause to `UPDATE` command

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #94 

**Related PRs**

- https://github.com/yiisoft/db-pgsql/pull/441
- https://github.com/yiisoft/db-sqlite/pull/386
- https://github.com/yiisoft/db-oracle/pull/359
- https://github.com/yiisoft/db-mssql/pull/392
- https://github.com/yiisoft/db-mysql/pull/422
- https://github.com/yiisoft/db-migration/pull/299
- https://github.com/yiisoft/active-record/pull/447

PR #1064, opened 2025-09-23
Fix: remove duplicated codes

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/db/pull/1038#discussion_r2296246677

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,7 +156,7 @@
 - New #1063: Add `indexBy()` method to `BatchQueryResultInterface` and `BatchQueryResult` class (@vjik)
 - Chg #1063: `AbstractConnection::createBatchQueryResult()` passes parameters `indexBy` and `resultCallback` to
   `BatchQueryResult` being created (@vjik)
-- Enh #1064: Remove duplicate code in favor of the `DbArrayHelper::normalizeToArray()` method (@rustamwin)
+- Enh #1064: Remove duplicate code in favor of the `DbArrayHelper::normalizeExpressions()` method (@rustamwin)
 
 ## 1.3.0 March 21, 2024
 


src/Helper/DbArrayHelper.php
```diff
@@ -287,6 +289,19 @@ public static function toArray(array|object $object): array
         return get_object_vars($object);
     }
 
+    public static function queryAsArray(array|ExpressionInterface|string $query): array
```
vjik: Seems, `$query` here is not query... I think another method and parameter names are required.

rustamwin: What you suggest? `value`/`expression`?

vjik: What is "$query" in this case?

rustamwin: The value after clauses (e.g FROM, SELECT, GROUP BY)

src/Helper/DbArrayHelper.php
```diff
@@ -289,16 +289,16 @@ public static function toArray(array|object $object): array
         return get_object_vars($object);
     }
 
-    public static function queryAsArray(array|ExpressionInterface|string $query): array
+    public static function normalizeToArray(array|ExpressionInterface|string $expression): array
```
vjik:
```suggestion
    /**
     * Normalizes raw input into an array of expression values.
     *
     * @param array|ExpressionInterface|string $raw Raw input to be normalized. It can be:
     *  - an array of expression values;
     *  - a single expression object;
     *  - a string with comma-separated expression values.
     *
     * @return array An array of normalized expressions.
     */
    public static function normalizeExpressions(array|ExpressionInterface|string $raw): array
```

vjik (approved):

vjik (approved):

PR #1076, opened 2025-10-21
Allow to use expressions as table name or condition in "join" query methods

### LINKED ISSUE #1073: Improve `Query::join()` methods
- parameter `array|string $table` should be extended to `array|ExpressionInterface|string $table` like for `Query::from()`
- parameter `array|string $on` should be extended to `array|ExpressionInterface|string $condition` like for `Query::where()`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
Fix #1073

### DIFF
--- a/src/QueryBuilder/AbstractDQLQueryBuilder.php
+++ b/src/QueryBuilder/AbstractDQLQueryBuilder.php
@@ -273,7 +273,7 @@ abstract class AbstractDQLQueryBuilder implements DQLQueryBuilderInterface
             $joins[$i] = "$joinType $table";
 
             if (isset($join[2])) {
-                if (is_array($join[2])) {
+                if (is_array($join[2]) && !isset($join[2][0])) {
                     foreach ($join[2] as &$column) {
                         if (is_string($column)) {
                             $column = new ColumnName($column);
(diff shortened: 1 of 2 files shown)


src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -256,28 +256,17 @@ public function buildJoin(array $joins, array &$params): string
...
-                    . 'condition.'
+                    'A join clause must be specified as an array of join type, join table, and optionally join condition.',
                 );
             }
 
-            /* 0:join type, 1:join table, 2:on-condition (optional) */
             [$joinType, $table] = $join;
 
-            $tables = $this->quoteTableNames((array) $table, $params);
+            $tables = $this->quoteTableNames(is_array($table) ? $table : [$table], $params);
```
Tigrov, line 269:
```suggestion
            $tables = $this->quoteTableNames((array) $table, $params);
```

This looks better

Tigrov (approved):
👍

src/QueryBuilder/AbstractDQLQueryBuilder.php
vjik: It's not work, because expression typecasting to array give property-value array.

PR #1079, opened 2025-10-23
Improve psalm types in `CommandInterface`  and `Query`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Command/AbstractCommand.php
+++ b/src/Command/AbstractCommand.php
@@ -547,10 +547,8 @@ abstract class AbstractCommand implements CommandInterface
 
         $this->setSql($sql)->bindValues($params);
 
-        /** @psalm-var array<string, mixed>|bool $result */
-        $result = $this->queryInternal(self::QUERY_MODE_ROW | self::QUERY_MODE_EXECUTE);
-
-        return is_array($result) ? $result : false;
+        /** @psalm-var array<string, mixed>|false */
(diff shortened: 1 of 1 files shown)


src/Command/AbstractCommand.php
```diff
@@ -395,7 +395,7 @@ public function insertReturningPks(string $table, array|QueryInterface $columns)
 
         $this->setSql($sql)->bindValues($params);
 
-        /** @psalm-var array|bool $result */
+        /** @psalm-var array<string, mixed>|bool $result */
         $result = $this->queryInternal(self::QUERY_MODE_ROW | self::QUERY_MODE_EXECUTE);
 
         return is_array($result) ? $result : false;
```
Tigrov:
```suggestion
        /** @psalm-var array<string, mixed>|false */
        return $this->queryInternal(self::QUERY_MODE_ROW | self::QUERY_MODE_EXECUTE);
```

src/Command/AbstractCommand.php
```diff
@@ -549,7 +549,7 @@ public function upsertReturning(
...
```
Tigrov:
```suggestion
        /** @psalm-var array<string, mixed>|false */
        return $this->queryInternal(self::QUERY_MODE_ROW | self::QUERY_MODE_EXECUTE);
```

Tigrov (approved):

PR #1080, opened 2025-10-24
Minor refactor `AbstractDQLQueryBuilder::buildColumns()` + Fix psalm types

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Query/QueryInterface.php
+++ b/src/Query/QueryInterface.php
@@ -30,7 +30,7 @@ use Yiisoft\Db\QueryBuilder\QueryBuilderInterface;
  * @psalm-type IndexBy = Closure(array|object):int|string
  * @psalm-type ResultCallback = Closure(non-empty-array<array>):non-empty-array<array|object>
  * @psalm-type ResultCallbackOne = Closure(array):(array|object)
- * @psalm-type JoinTable = array<string, ExpressionInterface|string>|ExpressionInterface|string
+ * @psalm-type JoinTable = array<ExpressionInterface|string>|ExpressionInterface|string
  * @psalm-type Join = list{string, JoinTable, array|ExpressionInterface|string}
  * @psalm-type From = array<string|ExpressionInterface>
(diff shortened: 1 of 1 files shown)


src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -160,19 +160,23 @@ public function buildColumns(array|string $columns): string
...
-        }
+        $columns = array_map(
+            function (string|ExpressionInterface $column): string {
+                if ($column instanceof ExpressionInterface) {
+                    return $this->buildExpression($column);
+                }
+                if (!str_contains($column, '(')) {
+                    return $this->quoter->quoteColumnName($column);
+                }
+                return $column;
```
Tigrov:
```suggestion
                return $this->quoter->quoteColumnName($column);
```

src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -160,19 +160,23 @@ public function buildColumns(array|string $columns): string
...
 
-        /** @psalm-var array<array-key, ExpressionInterface|string> $columns */
-        foreach ($columns as $i => $column) {
-            if ($column instanceof ExpressionInterface) {
-                $columns[$i] = $this->buildExpression($column);
-            } elseif (!str_contains($column, '(')) {
-                $columns[$i] = $this->quoter->quoteColumnName($column);
-            }
-        }
+        $columns = array_map(
```
Tigrov, line 167: Does `array_map(Closure, ...)` work faster than `foreach ()`?

Tigrov (approved):

src/QueryBuilder/AbstractDQLQueryBuilder.php
vjik: I don't know. But it's more clear for cases when we change source array, especially if short function syntax is used.

PR #1090, opened 2025-11-06
Improve README

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,11 @@ composer require yiisoft/db yiisoft/db-sqlite
 > [!IMPORTANT]
 > You must install `yiisoft/db` together with at least one database driver (e.g., `yiisoft/db-mysql`,
 > `yiisoft/db-pgsql`, `yiisoft/db-sqlite`, etc.) to actually connect to a database.
+>
+> It also depends on [PSR-16: Common Interface for Caching Libraries](https://www.php-fig.org/psr/psr-16/) and requires
+> the installation of [PSR-16 implementation](https://packagist.org/providers/psr/simple-cache-implementation).
+> For example, [yiisoft/cache](https://github.com/yiisoft/cache) or one of the other
+> [cache handlers](https://github.com/yiisoft/cache#cache-handlers).
 
 ## General Usage
 


README.md
```diff
@@ -7,36 +7,106 @@
...
+
+- PHP 8.1 - 8.4.
+- `pdo` PHP extension
+
+## Installation
+
+The package could be installed with [Composer](https://getcomposer.org).
+
+```shell
+composer require yiisoft/db yiisoft/db-sqlite
```
Tigrov, line 42:
```suggestion
composer require yiisoft/db-sqlite
```

README.md
```diff
@@ -7,36 +7,106 @@
...
+```
+
+> [!IMPORTANT]
+> You must install `yiisoft/db` together with at least one database driver (e.g., `yiisoft/db-mysql`,
+> `yiisoft/db-pgsql`, `yiisoft/db-sqlite`, etc.) to actually connect to a database.
```
Tigrov, line 47:
```suggestion
> You should install a database driver (e.g., `yiisoft/db-mysql`, `yiisoft/db-pgsql`, `yiisoft/db-sqlite`, etc.)
> to actually connect to a database. The `yiisoft/db` package will be installed as a dependency.
```

README.md
vjik: In this case `yiisoft/db` will be a transitive dependency, so it will be used in the project, but it will not be explicitly added to `composer.json`.

Tigrov: This is a bit different. DB requires a driver, and the driver depends on DB. They don't work without each other. If you remove the driver, you also remove DB.

vjik: The problem of transitive dependency is not only in existence of package, but also in the version.

User can write code for one Yii DB version, then update Yii DB Sqlite, Yii DB will be updated hidden, and user code can broken.

So transitive dependencies better always explicitly add to composer.json.

rustamwin (approved):

README.md
Tigrov: Will it not be a problem when updating the driver to a newer version that the DB package also needs to be updated?

vjik: > Will it not be a problem when updating the driver to a newer version that the DB package also needs to be updated?

Composer says to user about this and he will be able to update `yiisoft/db` and adapt own code, if required.

Tigrov: > User can write code for one Yii DB version, then update Yii DB Sqlite, Yii DB will be updated hidden, and user code can broken.

For me db with a driver looks like a single whole. If you update the driver, the db will of course also be updated. Specifying `yiisoft/db` in required section only adds complexity with that updates.

vjik: In my opinion, it is not worth promoting the use of transitive dependencies. And tools like `composer-require-checker` and `composer-dependency-analyser` will give this as an error.

Tigrov: If you think it is better - ok

PR #1113, opened 2025-11-30
PHP 8.5 support

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Related PRs:
- https://github.com/yiisoft/db-sqlite/pull/407
- https://github.com/yiisoft/db-mysql/pull/451
- https://github.com/yiisoft/db-mssql/pull/418
- https://github.com/yiisoft/db-oracle/pull/385
- https://github.com/yiisoft/db-pgsql/pull/469
- https://github.com/yiisoft/db-migration/pull/308
- https://github.com/yiisoft/active-record/pull/506

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,7 @@
 - Chg #1106: Remove parameters from `PdoConnectionInterface::getActivePdo()` method (@vjik)
 - Bug #1109: Fix column definition parsing in cases with brackets and escaped quotes (@vjik)
 - New #1107: Add abstract enumeration column type (@vjik)
+- Bug #1115: Do not reset `$retryHandler` property on `AbstractCommand::reset()` method call (@Tigrov)
 
 ## 1.3.0 March 21, 2024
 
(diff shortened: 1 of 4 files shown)


src/Schema/Column/IntegerColumn.php
```diff
@@ -46,6 +47,10 @@ public function phpTypecast(mixed $value): ?int
             return null;
         }
 
+        if (is_string($value)) {
+            $value = rtrim($value, '.');
+        }
+
```
Tigrov:
```suggestion
```

Is it required?

vjik: > Is it required?

MSSQL in edge cases pass `9223372036854775807.` as value and it throws warning in PHP 8.5.

Tigrov: Let's move it into `db-mssql` driver.

vjik: If user use this value, it also leads to unexpected warning. This may be in any DBMS.

vjik: Other side, it's PHP problem, not Yii. I'll try move it into `db-mssql`.

Tigrov: According to the doc "Converts the input value after retrieval from the database." we should check only values returned by DBMS. It is related with performance optimization.

PR #1114, opened 2025-11-30
Update doc for `CommandInterface::setRetryHandler()`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Command/CommandInterface.php
+++ b/src/Command/CommandInterface.php
@@ -770,9 +770,9 @@ interface CommandInterface
     public function setRawSql(string $sql): static;
 
     /**
-     * Sets a Closure (anonymous function) which called when a database exception is thrown when executing the command.
+     * Sets a closure (anonymous function) which called when a database exception is thrown when executing the command.
      *
-     * The signature of the Closure should be:
+     * The signature of the closure should be:
      *
      * ```php
      * use Yiisoft\Db\Exception\Exception;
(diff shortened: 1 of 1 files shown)


src/Command/CommandInterface.php
```diff
@@ -770,18 +770,24 @@ public function showDatabases(): array;
     public function setRawSql(string $sql): static;
 
     /**
-     * Sets a Closure (anonymous function) that's called when {@see Exception} is thrown when executing the
-     * command. The signature of the Closure should be:.
+     * Sets a Closure (anonymous function) which called when a database exception is thrown when executing the command.
```
vjik:
```suggestion
     * Sets a closure (anonymous function) which called when a database exception is thrown when executing the command.
```

src/Command/CommandInterface.php
```diff
@@ -770,18 +770,24 @@ public function showDatabases(): array;
...
+     *
+     * The signature of the Closure should be:
```
vjik:
```suggestion
     * The signature of the closure should be:
```

src/Command/CommandInterface.php
```diff
@@ -770,18 +770,24 @@ public function showDatabases(): array;
...
+     * function (Exception $e, int $attempt): bool
      * {
-     *     // return true or false (whether to retry the command or rethrow $e)
+     *     // return true or false (whether to retry the command or throw $e)
      * }
      * ```
      *
-     * The Closure will receive a database exception thrown and a current try (to execute the command) number
-     * starting from 1.
+     * The Closure will receive an {@see Exception} converted from the thrown database exception and the current attempt
```
vjik:
```suggestion
     * The closure will receive an {@see Exception} converted from the thrown database exception and the current attempt
```

src/Command/CommandInterface.php
```diff
@@ -770,18 +770,24 @@ public function showDatabases(): array;
...
+     * to execute the command, starting from `1`.
+     *
+     * If the Closure returns `true`, the command will be retried. If the Closure returns `false`, the {@see Exception}
```
vjik:
```suggestion
     * If the closure returns `true`, the command will be retried. If the closure returns `false`, the {@see Exception}
```

PR #1115, opened 2025-11-30
Do not reset `$retryHandler` property in `AbstractCommand::reset()`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | 

Related https://github.com/yiisoft/yii2/pull/19680


src/Command/AbstractCommand.php
```diff
@@ -639,7 +639,6 @@ protected function reset(): void
         $this->sql = '';
         $this->params = [];
         $this->refreshTableName = null;
-        $this->retryHandler = null;
```
vjik, line 642: Add test that checks `retryHandler` keeps its value.

PR #1116, opened 2025-11-30
Minor refactoring of `Query`

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -1,0 +1,72 @@
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - '.github/workflows/mutation.yml'
+      - 'composer.json'
+
+  push:
+    branches: ['master']
+    paths:
+      - 'src/**'
+      - '.github/workflows/mutation.yml'
+      - 'composer.json'
+
+name: Mutation
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  mutation:
+    name: PHP 8.5
+
+    env:
+      COMPOSER_ROOT_VERSION: dev-master
+
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:18
+        env:
+          POSTGRES_DB: yiitest
+          POSTGRES_USER: root
+          POSTGRES_PASSWORD: root
+        ports:
+          - 5432:5432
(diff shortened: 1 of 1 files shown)


src/Query/Query.php
```diff
@@ -152,16 +153,18 @@ public function addParams(array $params): static
...
+            return $this;
         }
 
+        if (is_int(key($params))) {
+            array_push($this->params, ...$params);
+
+            return $this;
+        }
+
+        $this->params = array_merge($this->params, $params);
```
samdark: Would this work?

```php
$this->params = [...$this->params, $params];
```

src/Query/Query.php
```diff
@@ -152,16 +153,18 @@ public function addParams(array $params): static
...
-                    $this->params[] = $value;
-                } else {
-                    $this->params[$name] = $value;
-                }
-            }
+
+            return $this;
         }
 
+        if (is_int(key($params))) {
```
vjik, line 160: According to interface method description, there are no integer keys:

```
@param array $params The list of query parameter values indexed by parameter placeholders.
     * For example, `[':name' => 'Dan', ':age' => 31]`.
```

If it's true, we should throw exception on int keys.
If it's false, we should update interface phpdoc.

Would it be a mixed array with both int and string keys?

Tigrov: Support for index parameters is available from previous versions.

It works somehow, but in some cases there will probably be errors due to the order of parameters not being preserved.

Removing this support seems a good idea. More opinions are needed for such a change.

src/Query/Query.php
Tigrov: Should work, looks like the same

samdark (approved):

src/Query/Query.php
vjik: PDO supports it, so we can keep it, but improve phpdoc.

Tigrov: If we add it in the doc we should support this. But currently it is somehow supported and no guarantees that there will be no errors. This can be used at own risk.

vjik: We can improve support and add more tests after major release, ok?

Tigrov: If this is necessary, we can create a separate issue for this

vjik: Yes. At least it can be discussed there.

Tigrov: Created an issue https://github.com/yiisoft/db/issues/1123

PR #1124, opened 2025-12-03
Add source of column information

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/408
- https://github.com/yiisoft/db-mysql/pull/453
- https://github.com/yiisoft/db-pgsql/pull/471
- https://github.com/yiisoft/db-mssql/pull/420
- https://github.com/yiisoft/db-oracle/pull/387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,7 +180,7 @@
 - Bug #1109: Fix column definition parsing in cases with brackets and escaped quotes (@vjik)
 - New #1107: Add abstract enumeration column type (@vjik)
 - Bug #1115: Do not reset `$retryHandler` property on `AbstractCommand::reset()` method call (@Tigrov)
-- Enh #1124: Add source of column information (@Tigrov)
+- New #1124: Add source of column information (@Tigrov)
 
 ## 1.3.0 March 21, 2024
 


src/Constant/ColumnInfoSource.php
```diff
@@ -0,0 +1,30 @@
...
+use Yiisoft\Db\Schema\Column\ColumnFactoryInterface;
+
+/**
+ * Defines the possible sources of column information.
+ *
+ * It can be used in `ColumnFactoryInterface` instances when creating column objects.
+ *
+ * @see ColumnFactoryInterface
+ */
+final class ColumnInfoSource
```
vjik, line 14: I suggest using an enumeration if the list of values is complete.

src/Schema/Column/AbstractColumnFactory.php
```diff
@@ -142,6 +143,7 @@ public function fromDefinition(string $definition, array $info = []): ColumnInte
         unset($definitionInfo['type']);
 
         $info = array_merge($info, $definitionInfo);
+        $info['source'] ??= ColumnInfoSource::DEFINITION;
```
vjik, line 146: Source is not "definition" into "fromDefinition" method, which feels strange.

src/Constant/ColumnInfoSource.php
Tigrov: It also can be user-defined sources like abstract types

src/Schema/Column/AbstractColumnFactory.php
Tigrov: Why not? Means that it is defined as a string and then parsed.

src/Schema/Column/ColumnFactoryInterface.php
```diff
@@ -26,14 +27,14 @@
...
  *     extra?: string|null,
- *     fromResult?: bool,
  *     primaryKey?: bool,
  *     name?: string|null,
  *     notNull?: bool,
  *     reference?: ForeignKey|null,
  *     scale?: int|null,
  *     schema?: string|null,
  *     size?: int|null,
+ *     source?: ColumnInfoSource::*,
```
vjik, line 37: Why is this info needed? It seems that `fromResult` and `source` are never used currently.

Tigrov: This can be useful when a user defines his own method/closure to create the column object in column factory.

PR #1128, opened 2025-12-06
Fix #1127 Schema->hasTable and Schema->hasView does not support yii-quoted names

### LINKED ISSUE #1127: Schema->hasTable and Schema->hasView does not support yii-quoted names
Schema->hasTable and Schema->hasView does not normalize table or view name before checking. 

### What steps will reproduce the problem?

1. Create table, let it be `example`;
2. Check `$db->schema->hasTable('{{%example}}');`
3. Check `$db->schema->hasTable('example');`

### What is the expected result?
`$db->schema->hasTable('{{%example}}') === true`
`$db->schema->hasTable('example') === true`

### What do you get instead?
`$db->schema->hasTable('{{%example}}') === false`
`$db->schema->hasTable('example') === true`

### Additional info

| Q                | A
| ---------------- | ---
| Version          | 2.0.0
| PHP version      |  8.4
| Operating system | Mac OS

batyrmastyr:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #1127

### DIFF
--- a/tests/Common/CommonSchemaTest.php
+++ b/tests/Common/CommonSchemaTest.php
@@ -283,7 +283,8 @@ abstract class CommonSchemaTest extends IntegrationTestCase
         $schema = $db->getSchema();
 
         $tempTableName = 'testTemporaryTable';
-        $db->createCommand()->dropTable($tempTableName, true)->execute();
+        $this->dropTable('testTemporaryTable');
+
         $db->createCommand()
             ->createTable(
                 $tempTableName,


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 2.0.1 under development
 
-- no changes in this release.
+- Bug #1127: Fix `AbstractSchema::hasTable` and `AbstractSchema::hasView` did not support yii-quoted `{{%table}}` and `{{%view}}` names (@batyrmastyr)
```
Tigrov:
```suggestion
- Bug #1127: Fix `AbstractSchema::hasTable()` and `AbstractSchema::hasView()` methods to support names quoted with curly brackets `{{%table}}` (@batyrmastyr)
```

tests/Common/CommonSchemaTest.php
```diff
@@ -283,6 +283,7 @@ public function testHasTable(): void
         $schema = $db->getSchema();
 
         $tempTableName = 'testTemporaryTable';
+        $db->createCommand()->dropTable($tempTableName, true)->execute();
```
vjik:
```suggestion
        $this->dropTable('testTemporaryTable');

```

Some drivers don't support "drop if exists".

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 2.0.1 under development
 
-- no changes in this release.
+- Bug #1127: Fix `AbstractSchema::hasTable()` and `AbstractSchema::hasView()` methods to support names quoted with curly brackets `{{%table}}` (@batyrmastyr)
```
vjik:
```suggestion
- Bug #1127: Fix `AbstractSchema::hasTable()` and `AbstractSchema::hasView()` methods to support names quoted with curly
  brackets `{{%table}}` (@batyrmastyr)
```

tests/Common/CommonSchemaTest.php
batyrmastyr: At this point table should not exist, except cases when test failed and normal drop was not executed.

vjik: > At this point table should not exist, except cases when test failed and normal drop was not executed.

Table not existing will lead to error in drivers that doesn't support "if exists".

Tigrov (approved):

PR #1144, opened 2026-01-02
Fix #1142: upsert to use first matching unique constraint

### LINKED ISSUE #1142: Upsert doesn't work when there's unique index
### Description

I've got the bug using PostgreSQL.

Create a table

```php
$b->createTable('page', [
    'id' => $cb::uuidPrimaryKey(),
    'title' => $cb::string()->notNull(),
    'slug' => $cb::string()->notNull()->unique(),
    'text' => $cb::text()->notNull(),
    'created_at' => $cb::dateTime(),
    'updated_at' => $cb::dateTime(),
]);
```

Do upsert:

```php
$this->connection->createCommand()->upsert('{{%page}}', [
    'id' => $page->id,
    'title' => $page->title,
    'slug' => $page->getSlug(),
    'text' => $page->text,
    'created_at' => $page->createdAt,
    'updated_at' => $page->updatedAt,
])->execute();
```

Get the error:

```
QLSTATE[42P10]: Invalid column reference: 7 ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification
The SQL being executed was: INSERT INTO "page" ("id", "title", "slug", "text", "created_at", "updated_at") VALUES ('019b5c77-1c8b-72bd-a0c9-547987f9cf9c', 'Test', 'test', 'test', '2025-12-26 21:01:08', '2025-12-26 21:01:08') ON CONFLICT ("id", "slug") DO UPDATE SET "title"=EXCLUDED."title", "text"=EXCLUDED."text", "created_at"=EXCLUDED."created_at", "updated_at"=EXCLUDED."updated_at"
```

### Package version

2.0.0

### PHP version

8.2

dadansatria:
| Q | A |
|---|---|
| Is bugfix? | ✔️ |
| New feature? | ❌ |
| Breaks BC? | ❌ |
| Fixed issues | #1142 |

When a table has multiple separate unique constraints (e.g., primary key on `id` and unique index on `slug`), the `upsert` method incorrectly merged all unique column names, generating invalid SQL like:

```sql
ON CONFLICT ("id", "slug") DO UPDATE ...
```

This failed because there's no composite unique constraint on `(id, slug)`.

**The fix:** Changed `getTableUniqueColumnNames()` to return columns from only the first matching constraint, instead of merging columns from all constraints.


dadansatria:
> Looks this will not work correct for some cases.
> 
> Better to generate expression for each unique index
> 
> Instead of `... ON CONFLICT ("id", "slug") DO UPDATE SET ...` should be `... ON CONFLICT ("id") DO UPDATE SET ... ON CONFLICT ("slug") DO UPDATE SET ... `

I checked the [PostgreeSQL docs](https://www.postgresql.org/docs/current/sql-insert.html) and unfortunately, multiple ON CONFLICT clauses are not allowed in a single INSERT statement

The syntax only allows ONE ON CONFLICT clause per INSERT

Tigrov:
> I checked the [PostgreeSQL docs](https://www.postgresql.org/docs/current/sql-insert.html) and unfortunately, multiple ON CONFLICT clauses are not allowed in a single INSERT statement
> 
> The syntax only allows ONE ON CONFLICT clause per INSERT

Seems this works for SQLite but not for PostgreSQL.

Then we should add an extra parameter to the `upsert()` method to specify the exact columns for the `ON CONFLICT (...)` clause.

samdark (approved):

src/QueryBuilder/AbstractDMLQueryBuilder.php
```diff
@@ -537,43 +545,29 @@ protected function getNormalizedColumnNames(array $columns): array
...
      * @param string[] $columns Source column list.
-     * @param Index[] $indexes This parameter optionally receives a matched index list.
-     * The constraints will be unique by their column names.
+     * @param Index[] $indexes This parameter optionally receives the first matched index,
+     * or an empty array if no matching index is found.
      *
-     * @return string[] The column names.
+     * @return string[] The column names of the first matching constraint.
     */
     private function getTableUniqueColumnNames(string $name, array $columns, array &$indexes = []): array
```
Tigrov, line 558:
```suggestion
    private function getTableUniqueColumnNames(string $name, array $columns, ?Index &$index = null): array
```

`array` can be changed to `?Index` with corresponding fixes of code

Tigrov (changes requested):
Dependent packages also need to be fixed (see test results)

Tigrov:
> 1. Fix `AbstractDMLQueryBuilder::getTableUniqueColumnNames()`. This is bugfix, and it doesn't break BC.

This probably doesn't solve the problem; see the issue #1142. This solution will select `id` field as a unique index, but the correct solution would be to specify the `slug` field for this.

> And one more idea... Configure globally default upsert constraint columns in connection. This is doesn't break BC and this will be enough in most cases.

This may also break BC if configure them via an additional method or parameter.

dadansatria:
Hi @vjik @Tigrov 

you're right, adding params to interface methods does break BC for custom implementations. How if I split this into two PRs:
1. **Bugfix only**: Fix `getTableUniqueColumnNames()` to use PK first, then first matching unique constraint
2. **New feature** (next minor): Add the `$constraintColumns` param
Let me know if this approach works. I'll start on the split..

PR #1163, opened 2026-02-21
Update installation doc

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

### Related PRs
- https://github.com/yiisoft/db-sqlite/pull/416
- https://github.com/yiisoft/db-mysql/pull/461
- https://github.com/yiisoft/db-pgsql/pull/485
- https://github.com/yiisoft/db-mssql/pull/427
- https://github.com/yiisoft/db-oracle/pull/394

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,15 @@ Optional packages that provide additional functionality:
 The package could be installed with [Composer](https://getcomposer.org).
 
 ```shell
-composer require yiisoft/db yiisoft/db-sqlite
+composer require yiisoft/db-sqlite:2.*
 ```
 
 > [!IMPORTANT]
-> You must install `yiisoft/db` together with at least one database driver (e.g., `yiisoft/db-mysql`,
-> `yiisoft/db-pgsql`, `yiisoft/db-sqlite`, etc.) to actually connect to a database.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,12 +41,15 @@ Optional packages that provide additional functionality:
 The package could be installed with [Composer](https://getcomposer.org).
 
 ```shell
-composer require yiisoft/db yiisoft/db-sqlite
+composer require yiisoft/db-sqlite
```
samdark: That's the instructions for `yiisoft/db` but we install `yiisoft/db-sqlite`?

I get the idea that one of the individual packages should be installed, and it will install `yiisoft/db` automatically. That, with links to individual packages, should be in this document, not an example about a single database.

samdark (changes requested):

README.md
Tigrov: It is not possible to install `yiisoft/db` without a concrete driver. But installing the concrete driver will also install `yiisoft/db`. So why specify `require yiisoft/db` if it doesn't matter?

samdark (approved):

PR #1165, opened 2026-02-23
Update workflows

Tigrov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/.github/workflows/active-record.yml
+++ b/.github/workflows/active-record.yml
@@ -1,16 +1,13 @@
 on:
   pull_request:
-    paths:
+    paths: &paths
       - 'src/**'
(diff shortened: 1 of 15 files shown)


vjik (changes requested):
Error with uploading code coverage:

```
error - 2026-03-11 12:56:00,259 -- Upload failed: {"message":"Token required because branch is protected"}
```

Example: https://github.com/yiisoft/db/actions/runs/22953429436/job/66623679217?pr=1165#step:8:349

I think we should clearly pass codecov token.

vjik (approved):

PR #1170, opened 2026-03-29
Fix #1169: Improve `ConvertException` error matching (#1170)

### LINKED ISSUE #1169: Error «no connection to server» is detected as Intergrity exception
### Description

When connection to MySQL server is lost you get
```
Yiisoft\Db\Exception\IntegrityException: SQLSTATE[HY000]: General error: 7 no connection to the server
``` 
instead of more correct `Yiisoft\Db\Exception\Exception` or, maybe, more specific, but not existing `ConnectionException`.

Source:
https://github.com/yiisoft/db/blob/1f38570d71d1555a3c81885b978493302e9c933f/src/Exception/ConvertException.php#L19-L21

### Package version

2.0

### PHP version

8.4

WarLikeLaux:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Docs added?   | ❌
| Tests added?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #1169

## What does this PR do?

Restricts `ConvertException` matching so generic `SQLSTATE[HY...]` errors are no longer treated as `IntegrityException`, `SQLSTATE[08...]` errors are converted to `ConnectionException`, Oracle integrity violations are recognized by dedicated `ORA-*` codes, and MySQL/MariaDB timeout disconnect errors plus Oracle `ORA-00942` keep the current compatibility path.

Adds unit coverage for message-based conversion and driver-suite coverage for foreign key violations and PDO command exception conversion.

No BC break. `ConnectionException` extends `Exception`, and the targeted compatibility carve-outs preserve current behavior for dependent packages in this CI matrix.

A full semantic cleanup would need coordinated changes in `yiisoft/db-mysql` and `yiisoft/db-migration`, which still rely on `IntegrityException` for reconnect and Oracle migration paths.

### DIFF
--- a/src/Exception/ConvertException.php
+++ b/src/Exception/ConvertException.php
@@ -18,6 +18,10 @@ final class ConvertException
 {
     private const MSG_CONNECTION_EXCEPTION = 'SQLSTATE[08';
     private const MSG_INTEGRITY_EXCEPTION = 'SQLSTATE[23';
+    private const MYSQL_RECONNECT_EXCEPTIONS = [
+        'SQLSTATE[HY000]: General error: 2006 ',
+        'SQLSTATE[HY000]: General error: 4031 ',
+    ];
     private const ORACLE_INTEGRITY_EXCEPTIONS = [
         'ORA-00001:',
         'ORA-01400:',
@@ -45,6 +49,7 @@ final class ConvertException
 
         if (
             str_contains($message, self::MSG_INTEGRITY_EXCEPTION)
+            || $this->isMysqlReconnectException($message)
             || $this->isOracleIntegrityException($message)
         ) {
             return new IntegrityException($message, $errorInfo, $this->e);
@@ -55,6 +60,17 @@ final class ConvertException
         }
 
         return new Exception($message, $errorInfo, $this->e);
+    }
+
+    private function isMysqlReconnectException(string $message): bool
+    {
+        foreach (self::MYSQL_RECONNECT_EXCEPTIONS as $mysqlReconnectException) {
+            if (str_contains($message, $mysqlReconnectException)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private function isOracleIntegrityException(string $message): bool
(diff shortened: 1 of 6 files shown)


tests/Db/Exception/ConvertExceptionTest.php
```diff
@@ -27,4 +38,57 @@ public function testRun(): void
...
+        ];
+    }
+
+    public static function integrityExceptionMessages(): array
+    {
+        return [
+            'mysql server has gone away' => ['SQLSTATE[HY000]: General error: 2006 MySQL server has gone away'],
+            'mysql server disconnected inactive client' => [
+                'SQLSTATE[HY000]: General error: 4031 The client was disconnected by the server because of inactivity.',
+            ],
```
Tigrov, line 83: These also should be converted to `ConnectionException`

WarLikeLaux: I agree, these cases should become `ConnectionException`. The problem is that the current CI matrix still installs `yiisoft/db-mysql` and `yiisoft/db-migration`, and they rely on the current `IntegrityException` behavior for these paths. If I switch `HY000 2006/4031` here now, the MySQL and MariaDB jobs fail. Could you please advise what would be the preferred way to proceed?

Tigrov: It requires to add changes in these packages too. The branch name should be the same

src/Exception/ConvertException.php
```diff
@@ -16,9 +16,23 @@
  */
 final class ConvertException
 {
-    private const MSG_INTEGRITY_EXCEPTION_1 = 'SQLSTATE[23';
-    private const MGS_INTEGRITY_EXCEPTION_2 = 'ORA-00001: unique constraint';
-    private const MSG_INTEGRITY_EXCEPTION_3 = 'SQLSTATE[HY';
+    private const MSG_CONNECTION_EXCEPTION = 'SQLSTATE[08';
+    private const MSG_INTEGRITY_EXCEPTION = 'SQLSTATE[23';
+    private const MYSQL_RECONNECT_EXCEPTIONS = [
```
bautrukevich, line 21: Why only MYSQL? It could be PostgreSQL also

Issue #1130, opened 2025-12-07
Need more environment for retry handler


Tigrov:
Needs to pass current connection `$this->db` to have possibility to restart it.

https://github.com/yiisoft/db/blob/f1074b91291e415b34059b85177a3e76c2bdd144/src/Driver/Pdo/AbstractPdoCommand.php#L210

PR #1171, opened 2026-03-29
Pass CommandInterface to retry handler (fixes #1155, #1130)

### LINKED ISSUE #1155: Renew connection if it was closed
### Proposed new feature or change

1. Collect params to bind 
    https://github.com/yiisoft/db/blob/e753995673aafcbd97e84a489ee09e0b80ee811d/src/Driver/Pdo/AbstractPdoCommand.php#L77
3. Pass `CommandInterface` (`$this`) to `retryHandler()` #1130 
    https://github.com/yiisoft/db/blob/e753995673aafcbd97e84a489ee09e0b80ee811d/src/Driver/Pdo/AbstractPdoCommand.php#L210
5. By default, try to renew connection and `pdoStatement` if the connection was closed
6. Throw an exception if a transaction was started

bautrukevich:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #1155, #1130

## Summary

This PR implements automatic connection recovery on the first execution failure and introduces a clean extension point via `CommandInterface` passed to the retry handler. It depends on PR #1170 (`ConnectionException`).

---

### New: `ConnectionRecoveryHandler`
All reconnect logic extracted into a standalone invokable class:
- Handles only `ConnectionException` (uses `instanceof`, no string matching)
- Skips reconnect if a transaction is active
- Calls `cancel()` + `prepare()` via `CommandInterface` to restore the statement on the new connection
- Set as the **default** `$this->retryHandler` in `AbstractPdoCommand::__construct()`

### Enh: `CommandInterface` passed to retry handler
Retry handler signature extended from `(Exception, int): bool` to `(Exception, int, CommandInterface): bool`.  
Old two-argument handlers continue to work — PHP silently ignores the extra argument.

### Enh: `bindPendingParams()` now covers `bindParam()` re-binding
`rebindBoundParams()` was removed; its loop is now the second `foreach` in `bindPendingParams()`.  
This means `bindParam()` bindings are automatically restored after `cancel()` + `prepare()` (e.g. on reconnect).

### Enh: Simplified `internalExecute()`
The hardcoded fallback reconnect block is gone. The loop is now a clean single dispatch:
```php
if ($this->retryHandler !== null && ($this->retryHandler)($e, $attempt, $this)) {
    continue;
}
throw $e;
```

---

## Breaking Changes

| What | Before | After |
|---|---|---|
| Retry handler signature | `(Exception, int): bool` | `(Exception, int, CommandInterface): bool` |
| Connection errors | May produce `IntegrityException` | Always produce `ConnectionException` |
| Default behaviour | No auto-reconnect | Auto-reconnects once on first `ConnectionException` |

Existing two-argument handlers keep working. Handlers that relied on catching `IntegrityException` for connection errors need to be updated to catch `ConnectionException`.

---

## Migration Guide

```php
// Before (still works, extra arg is ignored)
->setRetryHandler(fn(Exception $e, int $attempt) => $attempt < 3)

// After — can now access the command for custom logic
->setRetryHandler(function (Exception $e, int $attempt, CommandInterface $cmd): bool {
    if ($attempt === 0 && $e instanceof ConnectionException) {
        $cmd->cancel();
        // custom reconnect logic...
        $cmd->prepare();
        return true;
    }
    return $attempt < 3;
})

// To opt out of the default auto-reconnect entirely:
->setRetryHandler(null)
```

---

## Usage Example

```php
// Default behaviour (no code needed — built in):
// On first ConnectionException → close + reopen + re-prepare → retry once.

// Custom handler that retries up to 3 times and logs:
$command->setRetryHandler(
    function (Exception $e, int $attempt, CommandInterface $cmd) use ($logger): bool {
        $logger->warning("DB error attempt $attempt: {$e->getMessage()}");
        return $attempt < 3;
    }
);
```

```

### CODE REVIEW

File: src/Command/AbstractCommand.php [hunk 1]
```diff
@@ -472,6 +472,30 @@ public function setSql(string $sql): static
...
+     * ```
+     *
+     * The closure will receive an {@see Exception} converted from the thrown database exception,
+     * the current attempt to execute the command (starting from `0`), and the {@see CommandInterface}
+     * instance to allow access to the command's parameters and other state exposed by the interface.
+     *
+     * If the closure returns `true`, the command will be retried. If the closure returns `false`,
+     * the {@see Exception} will be thrown.
+     *
+     * @param Closure|null $handler A PHP callback to handle database exceptions.
```
samdark: I think `@psalm-param` with function signature could be added.

src/Command/CommandInterface.php
```diff
@@ -776,18 +776,20 @@ public function setRawSql(string $sql): static;
...
+     * The closure will receive an {@see Exception} converted from the thrown database exception,
+     * the current attempt to execute the command (starting from `0`), and the {@see CommandInterface}
+     * instance to allow access to the command and its parameters for custom retry logic.
      *
-     * If the closure returns `true`, the command will be retried. If the closure returns `false`, the {@see Exception}
-     * will be thrown.
+     * If the closure returns `true`, the command will be retried. If the closure returns `false`,
+     * the {@see Exception} will be thrown.
      *
      * @param Closure|null $handler A PHP callback to handle database exceptions.
```
samdark, line 794: I think `@psalm-param` with function signature could be added.

src/Command/AbstractCommand.php
```diff
@@ -472,6 +472,31 @@ public function setSql(string $sql): static
...
+     * @psalm-param Closure(Exception, int, CommandInterface): bool|null $handler
+     */
```
Tigrov:
```suggestion
```

These annotations are not needed if they are the same as in the interface `CommandInterface`

src/Driver/Pdo/AbstractPdoCommand.php
Tigrov: Can it be moved to `bindPendingParams()`?

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -174,6 +187,35 @@ protected function bindPendingParams(): void
...
+            } elseif ($entry['driverOptions'] === null) {
+                $this->pdoStatement?->bindParam($name, $value, $entry['type'], $entry['length']);
+            } else {
+                $this->pdoStatement?->bindParam($name, $value, $entry['type'], $entry['length'], $entry['driverOptions']);
+            }
+
+            unset($value);
+        }
+        unset($entry);
+    }
```
Tigrov: Can it be moved to `bindPendingParams()`?

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -207,9 +252,37 @@ protected function internalExecute(): void
...
+                        $this->pdoStatement = null;
+                    } catch (Throwable) {
+                        // If reconnection fails, throw original error
+                        throw $e;
+                    }
+
+                    // Re-prepare the statement against the new connection, restoring all parameter bindings.
+                    $this->prepare();
+                    continue; // Retry the command
+                }
```
Tigrov: Looks better to move this code to separate class `ConnectionRecoveryHandler` and set it by default to `$this->retryHandler`

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -309,6 +382,27 @@ protected function queryInternal(int $queryMode): mixed
...
+    {
+        $message = $e->getMessage();
+
+        return str_contains($message, 'no connection')
+            || str_contains($message, 'General error: 7')
+            || str_contains($message, 'gone away')
+            || str_contains($message, 'Connection refused')
+            || str_contains($message, 'server has gone away')
+            || str_contains($message, 'Lost connection');
+    }
```
Tigrov: Looks like this part depends on another PR
https://github.com/yiisoft/db/pull/1170

And could be replaced with `$e instanceof ConnectionException`

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -87,6 +94,11 @@ public function bindParam(
             $dataType = $this->db->getSchema()->getDataType($value);
         }
 
+        // Save the binding by reference so it can be re-applied after statement re-preparation (e.g., on reconnect).
+        $entry = ['type' => $dataType, 'length' => $length, 'driverOptions' => $driverOptions, 'value' => null];
+        $entry['value'] = &$value;
```
vjik:
```suggestion
        $entry = ['value' => &$value, 'dataType' => $dataType, 'length' => $length, 'driverOptions' => $driverOptions];
```

For consistency with `bindParam()` signature.

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -154,6 +166,7 @@ public function prepare(?bool $forRead = null): void
         try {
             $this->pdoStatement = $pdo->prepare($sql);
             $this->bindPendingParams();
+            $this->rebindBoundParams();
```
vjik: Now the parameters are binded twice: on call `bindParam()` and here. Can we rebind only after reconnect?

vjik (changes requested):
Add line to changelog

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -53,12 +53,20 @@ abstract class AbstractPdoCommand extends AbstractCommand implements PdoCommandI
...
+     */
+    protected array $pendingBoundParams = [];
+
     /**
      * @param PdoConnectionInterface $db The PDO database connection to use.
      */
     public function __construct(PdoConnectionInterface $db)
     {
         parent::__construct($db);
+        $this->retryHandler = (new ConnectionRecoveryHandler($db))->asClosure();
```
Tigrov, line 69:
```suggestion
        $this->retryHandler = (new ConnectionRecoveryHandler())(...);
```

Is there way to get db connection from the command?

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -81,19 +89,16 @@ public function bindParam(
...
-        }
+        $this->pendingBoundParams[$name] = [
+            'type' => $dataType,
+            'length' => $length,
+            'driverOptions' => $driverOptions,
+            'value' => &$value,
+        ];
 
         return $this;
     }
```
Tigrov, line 104: Now this method does not depend on `pdoStatement`, we can move it to the `AbstractCommand` class.

src/Driver/Pdo/AbstractPdoCommand.php
```diff
@@ -53,12 +53,20 @@ abstract class AbstractPdoCommand extends AbstractCommand implements PdoCommandI
      */
     protected ?PDOStatement $pdoStatement = null;
 
+    /**
+     * @var array<int|string, array{value: mixed, type: int, length: int|null, driverOptions: mixed}>
+     * Parameters bound via {@see bindParam()} stored by reference for re-binding after statement re-preparation
+     * (e.g., on reconnect).
+     */
+    protected array $pendingBoundParams = [];
```
Tigrov, line 61:
```suggestion
    protected array $bindingParams = [];
```

We have `AbstractCommand::$params`, which in PDO is bound using `PDOStatement::bindValue()`.
And `$pendingBoundParams` is bound using `PDOStatement::bindParam()`.
I suggest to rename it to `$bindingParams` or `$pendingParams`.

Perhaps good idea to rename `AbstractCommand::$params` to `$bindingValues` or `$pendingValues` in the future.

src/Driver/Pdo/ConnectionRecoveryHandler.php
```diff
@@ -0,0 +1,66 @@
...
+        return true;
+    }
+
+    /**
+     * @psalm-return RetryHandlerClosure
+     */
+    public function asClosure(): Closure
+    {
+        return $this->__invoke(...);
+    }
```
Tigrov, line 65:
```suggestion
```

Use `(new ConnectionRecoveryHandler())(...)`

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## 2.0.2 under development
 
-- no changes in this release.
+- New #1170: Add `ConnectionException` class for connection-related errors; `ConvertException` now returns it instead of plain `Exception` for lost/refused connections (@your-username)
+- New #1171: Add `ConnectionRecoveryHandler` class; automatic reconnect on first connection error is now the default retry handler; pass `CommandInterface` as third argument to retry handler callback (@your-username)
+- Enh #1171: Merge `rebindBoundParams()` into `bindPendingParams()`; `bindParam()` bindings are now restored after reconnect (@your-username)
```
Tigrov, line 7:
```suggestion
- New #1171: Add `ConnectionRecoveryHandler` class; automatic reconnect on first connection error is now the default retry handler; pass `CommandInterface` as third argument to retry handler callback (@bautrukevich)
- Enh #1171: Bind params are now collected and bound before execution (@bautrukevich)
```

PR #1173, opened 2026-04-02
Fix buildLimit() silently dropping OFFSET 0

darkspock:
## Summary

`AbstractDQLQueryBuilder::buildLimit()` uses `!empty($offset)` to decide whether to append the OFFSET clause. Since `empty(0)` returns `true` in PHP, an explicitly set `OFFSET 0` is silently dropped from the generated SQL.

While `OFFSET 0` has no semantic effect on query results, this is inconsistent: a value explicitly set by the caller should be reflected in the output. It can also cause confusion when debugging generated SQL.

### Fix

Replace `!empty($offset)` with `$offset !== null`, which correctly distinguishes "no offset set" (`null`) from "offset is zero" (`0`).

### Before
```php
if (!empty($offset)) {
```

### After
```php
if ($offset !== null) {
```

## Risk

Very low — the only change is that `OFFSET 0` now appears in the SQL when explicitly set. This has no effect on query results.

### DIFF
--- a/src/QueryBuilder/AbstractDQLQueryBuilder.php
+++ b/src/QueryBuilder/AbstractDQLQueryBuilder.php
@@ -300,7 +300,7 @@ abstract class AbstractDQLQueryBuilder implements DQLQueryBuilderInterface
                 . ($limit instanceof ExpressionInterface ? $this->buildExpression($limit) : (string) $limit);
         }
 
-        if ($offset !== null) {
+        if (!empty($offset)) {
             $sql .= ' OFFSET '
                 . ($offset instanceof ExpressionInterface ? $this->buildExpression($offset) : (string) $offset);
         }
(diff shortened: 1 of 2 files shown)


src/QueryBuilder/AbstractDQLQueryBuilder.php
```diff
@@ -300,7 +300,7 @@ public function buildLimit(ExpressionInterface|int|null $limit, ExpressionInterf
                 . ($limit instanceof ExpressionInterface ? $this->buildExpression($limit) : (string) $limit);
         }
 
-        if (!empty($offset)) {
+        if ($offset !== null) {
```
Tigrov: I don't think this change is necessary. At least if there is no real case where it is needed.

vjik (changes requested):
"no offset set" (null) and "offset is zero" (0) gives equal result in SQL. I agree with @Tigrov and suggest don't change it.

But it will be useful to add an explanation to the description of methods that set offset.

src/QueryBuilder/AbstractDQLQueryBuilder.php
Tigrov:
```suggestion
        if (!empty($offset)) {
```

Revert changes

Tigrov (approved):

vjik (approved):

PR #1174, opened 2026-04-02
Fix ConvertException false positives from SQLSTATE[HY match

darkspock:
## Summary

`ConvertException::run()` matches `SQLSTATE[HY` in error messages to detect integrity constraint violations. However, SQLSTATE class `HY` is the "CLI-specific condition" class (e.g., `HY000` = general error, `HY001` = memory allocation error, `HY010` = function sequence error). These are **not** integrity constraint violations and should not be wrapped as `IntegrityException`.

This can cause false positives where general database errors are incorrectly classified as integrity violations.

### Fix

- Remove the `SQLSTATE[HY` string match
- Add a proper check using `PDOException::errorInfo[0]` (the SQLSTATE code) for class `23`, which is the standard SQLSTATE class for integrity constraint violations across all DBMS
- Keep the existing `SQLSTATE[23` message string match and `ORA-00001` match as fallbacks

### Before
```php
return match (
    str_contains($message, self::MSG_INTEGRITY_EXCEPTION_1)
    || str_contains($message, self::MGS_INTEGRITY_EXCEPTION_2)
    || str_contains($message, self::MSG_INTEGRITY_EXCEPTION_3) // SQLSTATE[HY — too broad
) {
```

### After
```php
$isIntegrity = str_contains($message, self::MSG_INTEGRITY_EXCEPTION_1)
    || str_contains($message, self::MGS_INTEGRITY_EXCEPTION_2)
    || $this->isIntegrityConstraintViolation($errorInfo); // checks errorInfo[0] starts with '23'
```

## Risk

Low — this makes detection **more precise**. Errors that were previously (incorrectly) classified as `IntegrityException` will now correctly be classified as `Exception`. Code that catches `IntegrityException` specifically will no longer catch unrelated errors.

### DIFF
--- a/src/Exception/ConvertException.php
+++ b/src/Exception/ConvertException.php
@@ -18,7 +18,6 @@ final class ConvertException
 {
     private const MSG_INTEGRITY_EXCEPTION_1 = 'SQLSTATE[23';
     private const MGS_INTEGRITY_EXCEPTION_2 = 'ORA-00001: unique constraint';
-    private const MSG_INTEGRITY_EXCEPTION_3 = 'SQLSTATE[HY';
 
     public function __construct(
         private readonly \Exception $e,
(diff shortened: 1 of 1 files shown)


Tigrov:
Another PR already fixing this part of code.
https://github.com/yiisoft/db/pull/1170

If you have any suggestions regarding this, please write comment in that PR

vjik:
Closed in favor #1170

PR #1196, opened 2026-08-19
New #868: Add SerializationFailureException

### LINKED ISSUE #868: Add SerializationFailureException
Add SerializationFailureException.

SQLSTATE code 40001
PDO exception substring: SQLSTATE[40001]

Code is the same for all databases (mysql, pgsql, mssql, oracle).

KalimeroMK:
Fix #868

Adds `SerializationFailureException` for SQLSTATE code 40001 (serialization failure, e.g. transaction deadlock in concurrent updates). The code is the same for all databases, so the conversion is done centrally in `ConvertException`, next to the existing `IntegrityException` mapping.


Tigrov (approved):

src/Exception/ConvertException.php
```diff
@@ -36,6 +37,10 @@ public function run(): Exception
 
         $errorInfo = $this->e instanceof PDOException ? $this->e->errorInfo : null;
 
+        if (str_contains($message, self::MSG_SERIALIZATION_FAILURE)) {
```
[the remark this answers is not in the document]
vjik: Also tests with SQL that contains "SQLSTATE[40001]" is needed.

vjik: What about check ` $errorInfo[0] === 40001` ([docs](https://www.php.net/manual/pdo.errorinfo.php#refsect1-pdo.errorinfo-returnvalues))?

src/Exception/ConvertException.php
vjik: Actualize this phpdoc

vjik (changes requested):

src/Exception/ConvertException.php
```diff
@@ -36,6 +38,13 @@ public function run(): Exception
 
         $errorInfo = $this->e instanceof PDOException ? $this->e->errorInfo : null;
 
+        if (
+            ($errorInfo[0] ?? null) === '40001'
+            || str_contains($this->e->getMessage(), self::MSG_SERIALIZATION_FAILURE)
```
vjik: Are there cases when `($errorInfo[0] ?? null) === '40001'` is not enough?

zoranbogoevskimkd: in real usage PDO always populates errorInfo, so the message fallback only served manually constructed exceptions I Removed it so  the check now relies solely on $errorInfo[0] and the test sets errorInfo the  way PDO would.

src/Connection/AbstractConnection.php
```diff
@@ -32,9 +32,7 @@ public function beginTransaction(?string $isolationLevel = null): TransactionInt
         $this->open();
         $this->transaction = $this->getTransaction();
 
-        if ($this->transaction === null) {
-            $this->transaction = $this->createTransaction();
-        }
+        $this->transaction ??= $this->createTransaction();
```
Tigrov: Can be merged into `$this->transaction = $this->getTransaction() ?? $this->createTransaction();`

rector.php
```diff
@@ -34,6 +35,9 @@
             __DIR__ . '/tests/Common/CommonColumnTest.php',
             __DIR__ . '/tests/Db/Schema/Column/ColumnTest.php',
         ],
+        IfToNullCoalescingAssignRector::class => [
+            __DIR__ . '/src/Driver/Pdo/PdoServerInfo.php',
+        ],
```
Tigrov: Why is this an exception?

Looks like it should also be converted

https://github.com/yiisoft/db/blob/e7d19eb05b9ddd3f9793ce0869aef4bcf4439906/src/Driver/Pdo/PdoServerInfo.php#L24-L27

KalimeroMK: @Tigrov pr update

src/Driver/Pdo/PdoServerInfo.php
```diff
@@ -21,11 +23,13 @@ public function getTimezone(bool $refresh = false): string
 
     public function getVersion(): string
     {
-        if ($this->version === null) {
-            /** @var string */
-            $this->version = $this->db->getActivePdo()->getAttribute(PDO::ATTR_SERVER_VERSION) ?? '';
-        }
+        return $this->version ??= $this->fetchVersion();
```
Tigrov:
```suggestion
        return $this->version ??= (string) $this->db->getActivePdo()->getAttribute(PDO::ATTR_SERVER_VERSION);
```
This looks quite simpler.

tests/Db/Driver/Pdo/PdoCommandTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+        $params = $command->getParams(false);
+
+        $this->assertSame(DataType::INTEGER, $params[':int']->type);
+        $this->assertSame(DataType::STRING, $params[':string']->type);
+        $this->assertSame(DataType::NULL, $params[':null']->type);
+        $this->assertSame(DataType::STRING, $params[':explicit']->type);
+
+        $db->close();
+    }
```
Tigrov: Not sure these tests are relevant to the issue. If add the tests they should not be for SQLite only, but for other DBMS also. It looks good to improve `\Yiisoft\Db\Tests\Common\CommonPdoCommandTest::testBindParam()` method or move there tests into `CommonPdoCommandTest` class for all supported PDO DBMS.

src/Driver/Pdo/PdoServerInfo.php
KalimeroMK: Applied. Psalm types `PDO::getAttribute()` as possibly returning an array, so the cast needs a targeted suppression:

```php
/** @psalm-suppress PossiblyInvalidCast Psalm types PDO::getAttribute() as possibly returning an array. */
return $this->version ??= (string) $this->db->getActivePdo()->getAttribute(PDO::ATTR_SERVER_VERSION);
```

Without it: `PossiblyInvalidCast - array<array-key, mixed> cannot be cast to string`. The `rector.php` exception is gone.

tests/Db/Driver/Pdo/PdoCommandTest.php
KalimeroMK: You're right, they weren't relevant — removed.

I added them because Codecov reported three lines as missing, but those lines are already covered for every supported DBMS:

- `AbstractPdoCommand::bindParam()` — `CommonPdoCommandTest` L85-87, L94 call it without an explicit `$dataType`
- `AbstractPdoCommand::bindValue()` — `CommonPdoCommandTest` L104, L111
- `AbstractConnection::beginTransaction()` — `CommonPdoConnectionTest` L131, L187, L244, L281 and `CommonConnectionTest` L59 (the nested/reuse branch)

Tigrov (approved):

vjik (approved):

PR #1197, opened 2026-08-20
Chg #960: Pass non-null and non-expression values only to dbTypecast()

### LINKED ISSUE #960: Pass to `ColumnInterface::dbTypecast()` non-null and non-expression values only
Seems, all implementations of `ColumnInterface::dbTypecast()`  checks value on `null` and `ExpressionInterface` and return it as is for those values:

```php
if ($value === null || $value instanceof ExpressionInterface) {
    return $value;
}
```

We can check it before call `dbTypecast()` and simplify simplify both existing implementations and the creation of custom ones.

KalimeroMK:
Fix #960

## Changes

- `null` values are now passed through without calling `ColumnInterface::dbTypecast()`. The check is inlined at the call sites in `AbstractDMLQueryBuilder` (insert/update/batch) and `AbstractColumnDefinitionBuilder::buildDefault()`.
- No new API. The first version of this PR added a `dbTypecast()` wrapper function, but a wrapper that only filters `null` adds call overhead for no measurable gain (see benchmark), so the check was inlined instead.
- `DateTimeValueBuilder` is untouched: `DateTimeValue::$value` is non-nullable by type.
- Column implementations are intentionally left untouched. Two findings from the driver test matrices:
  - `Expression` is `Stringable`, so removing the `ExpressionInterface` arms silently stringifies expressions instead of passing them through (breaks DDL builders in db-pgsql/db-mssql that pass `Expression` default values directly).
  - Some driver implementations process expressions: `Yiisoft\Db\Mssql\Column\BinaryColumn::dbTypecast()` unwraps `Param` values into `CONVERT(VARBINARY(MAX), ...)` expressions, so expressions must keep reaching `dbTypecast()`.

  Net effect: only `null` is safe to pre-filter. Simplifying the implementations needs a coordinated change across the driver packages and can be done as a follow-up for 3.0.

## Benchmark

2M calls per round, median of 7 rounds, PHP 8.5 with OPcache + JIT (tracing), workload: 80% scalars, 10% `null`, 10% expressions, `IntegerColumn` + `StringColumn`:

| approach | time per round |
|---|---|
| current master (direct call) | ~34 ms |
| wrapper function (first version of this PR) | ~42 ms |
| check inlined at the call site (this PR) | ~34.5 ms |

The inlined check is within noise of master; the wrapper function added ~4 ns per call on trivial casts.


Tigrov:
But it seems the function that only checks for `null` doesn't make sense.

https://github.com/yiisoft/db/issues/960#issuecomment-2835306541

KalimeroMK:
Agreed that a null-only wrapper doesn't pay for itself — dropped the function and inlined the null check at the call sites in 82017d68. Benchmark is at parity with master now.

Tigrov:
I'd leave it as is and close the issue without changes. The PR found that `Expression` cannot be moved outside of the type caster of the column classes. We have custom type casts possible, and `null` can be handled as a special case as described in the comment https://github.com/yiisoft/db/issues/960#issuecomment-2816844677

samdark:
Yes, I agree that in this case the repetition is unavoidable and is actually okay.

samdark:
@KalimeroMK thank you for the pull request. It helped.

Issue #1152, opened 2026-01-11
A unversal way to work with UUIDs


samdark:
### Proposed new feature or change

Currently you can't work universally with UUIDs:

```php
<?php

declare(strict_types=1);

namespace App\Migration;

use Yiisoft\Db\Migration\MigrationBuilder;
use Yiisoft\Db\Migration\RevertibleMigrationInterface;

final class M251102141707Page implements RevertibleMigrationInterface
{
    public function up(MigrationBuilder $b): void
    {
        $column = $b->columnBuilder();

        $b->createTable('page', [
            'id' => $column::uuidPrimaryKey(),
            'title' => $column::string()->notNull(),
            'slug' => $column::string()->notNull()->unique(),
            'text' => $column::text()->notNull(),
            'created_at' => $column::dateTime(),
            'updated_at' => $column::dateTime(),
        ]);
    }

    public function down(MigrationBuilder $b): void
    {
        $b->dropTable('page');
    }
}
```

Creates a proper column for PostgreSQL and MySQL but then if you want to insert UUID, there's no universal way to do so:

```php
$this->connection->createCommand()->insert('{{%page}}', [
            'id' => Uuid::uuid7()->getBytes(), // MySQL
            // 'id' => Uuid::uuid7()->toString(), // PostgreSQL
            'title' => 'title',
            'slug' => 'slug',
            'text' => 'text',
            'created_at' => new DateTimeImmutable(),
            'updated_at' => new DateTimeImmutable(),
        ])->execute();
```

MySQL accepts bytes and errors with strings. PostgreSQL accepts strings and errors with bytes.

samdark:
https://github.com/yiisoft/docs/pull/324

Tigrov:
One more reason why `UuidColumn` class is required. At least in PostgreSQL it can replace bytes to string.

PR #1199, opened 2026-09-07
New #1152: Add `UuidValue` expression for DBMS-independent UUID values

KalimeroMK:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #1152

Adds `UuidValue`, an expression that carries a UUID and is bound in the representation the current DBMS expects — raw bytes for MySQL, MariaDB, SQLite and Oracle, canonical string for PostgreSQL and MSSQL:

```php
$db->createCommand()->insert('{{%page}}', ['id' => new UuidValue(Uuid::uuid7())])->execute();
```

It accepts the canonical form, 32 hexadecimal characters, 16 raw bytes or any `Stringable`. Passing raw bytes or a raw string directly keeps working as before.

### DIFF
--- a/src/Expression/Value/Builder/UuidValueBuilder.php
+++ b/src/Expression/Value/Builder/UuidValueBuilder.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Db\Expression\Value\Builder;
+
+use Yiisoft\Db\Expression\ExpressionBuilderInterface;
+use Yiisoft\Db\Expression\ExpressionInterface;
+use Yiisoft\Db\Expression\Value\UuidValue;
+use Yiisoft\Db\QueryBuilder\QueryBuilderInterface;
+use Yiisoft\Db\Helper\DbUuidHelper;
+
+/**
+ * Builder for {@see UuidValue} expressions.
+ *
+ * Binds the UUID in the canonical string form, which is what PostgreSQL `uuid` and MSSQL `uniqueidentifier` columns
+ * expect. DBMS that store a UUID as raw bytes, such as MySQL, MariaDB, SQLite and Oracle, override
+ * {@see prepareValue()} to convert the value with {@see DbUuidHelper::uuidToBlob()}.
+ *
+ * @implements ExpressionBuilderInterface<UuidValue>
+ */
+class UuidValueBuilder implements ExpressionBuilderInterface
+{
+    /**
+     * @param QueryBuilderInterface $queryBuilder The query builder instance.
+     */
+    public function __construct(
+        protected readonly QueryBuilderInterface $queryBuilder,
+    ) {}
+
+    public function build(ExpressionInterface $expression, array &$params = []): string
+    {
+        return $this->queryBuilder->buildValue($this->prepareValue($expression), $params);
+    }
+
+    /**
+     * Converts the UUID to the representation expected by the DBMS.
+     *
+     * @param UuidValue $expression The expression to convert.
+     *
+     * @return mixed The value to bind, it's passed to {@see QueryBuilderInterface::buildValue()}.
+     */
+    protected function prepareValue(UuidValue $expression): mixed
+    {
+        return $expression->value;
+    }
+}
--- a/src/Expression/Value/Builder/UuidValueBuilder.php
+++ b/src/Expression/Value/Builder/UuidValueBuilder.php
@@ -4,18 +4,21 @@ declare(strict_types=1);
 
 namespace Yiisoft\Db\Expression\Value\Builder;
 
+use Yiisoft\Db\Constant\DataType;
 use Yiisoft\Db\Expression\ExpressionBuilderInterface;
 use Yiisoft\Db\Expression\ExpressionInterface;
+use Yiisoft\Db\Expression\Value\Param;
 use Yiisoft\Db\Expression\Value\UuidValue;
-use Yiisoft\Db\QueryBuilder\QueryBuilderInterface;
 use Yiisoft\Db\Helper\DbUuidHelper;
+use Yiisoft\Db\QueryBuilder\QueryBuilderInterface;
 
 /**
  * Builder for {@see UuidValue} expressions.
  *
- * Binds the UUID in the canonical string form, which is what PostgreSQL `uuid` and MSSQL `uniqueidentifier` columns
- * expect. DBMS that store a UUID as raw bytes, such as MySQL, MariaDB, SQLite and Oracle, override
- * {@see prepareValue()} to convert the value with {@see DbUuidHelper::uuidToBlob()}.
+ * Binds the UUID as a string parameter in the canonical form, which is what PostgreSQL `uuid` and MSSQL
+ * `uniqueidentifier` columns expect. DBMS that store a UUID as raw bytes, such as MySQL, MariaDB, SQLite and Oracle,
+ * override {@see prepareValue()} to convert the value with {@see DbUuidHelper::uuidToBlob()} and bind it as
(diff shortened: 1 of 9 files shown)


src/Expression/Value/Builder/UuidValueBuilder.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @param QueryBuilderInterface $queryBuilder The query builder instance.
+     */
+    public function __construct(
+        protected readonly QueryBuilderInterface $queryBuilder,
+    ) {}
+
+    public function build(ExpressionInterface $expression, array &$params = []): string
+    {
+        return $this->queryBuilder->buildValue($this->prepareValue($expression), $params);
```
Tigrov, line 36: If UUID is binary value, should it be built as binary?

src/Expression/Value/UuidValue.php
```diff
@@ -0,0 +1,62 @@
...
+    public function __construct(string|Stringable $value)
+    {
+        try {
+            $this->value = strtolower(DbUuidHelper::toUuid((string) $value));
+        } catch (InvalidArgumentException $e) {
+            throw new InvalidArgumentException(
+                'Value is not a valid UUID. Expected the canonical form, 32 hexadecimal characters or 16 raw bytes.',
+                previous: $e,
+            );
+        }
```
Tigrov: Suggest to change the message of current exception instead adding one more try-catch.

https://github.com/yiisoft/db/blob/11032c2eb6fe5971538a3d63ab14431e502d37a8/src/Helper/DbUuidHelper.php#L28

Tigrov (commented):
This looks like a part of realization, with canonical string form. What about realizations for other concrete DBMS?

src/Expression/Value/Builder/UuidValueBuilder.php
KalimeroMK: Fixed in 6795ad2. `prepareValue()` now returns a `Param`, so DBMS that store a UUID as raw bytes can bind it as `DataType::LOB` instead of letting `buildValue()` infer `DataType::STRING`.

src/Expression/Value/UuidValue.php
KalimeroMK: Fixed in 6795ad2. Reworded the exception in `DbUuidHelper::toUuid()` and dropped the extra try/catch in `UuidValue`.

KalimeroMK:
@Tigrov correct, this is only the core half — the default builder covers PostgreSQL and MSSQL, while MySQL/MariaDB/SQLite/Oracle override `prepareValue()` to bind the raw bytes.

Correction to what I wrote earlier about CI: no tag is needed. `install-packages` resolves `yiisoft/db` from a same-named branch in my fork, so I can open the driver PRs from `add-1152-uuid-value` and they will be green against this branch right away. I will do that, so you can review the whole thing as one unit.

KalimeroMK:
@Tigrov driver PRs are up, green against this branch on real databases:

- yiisoft/db-sqlite#432 — `blob(16)`
- yiisoft/db-mysql#482 — `binary(16)`, MariaDB included
- yiisoft/db-oracle#411 — `raw(16)`

`db-pgsql` and `db-mssql` need nothing.

Oracle could not use the `prepareValue()` seam: PDO_OCI inserts `NULL` for a `raw(16)` column bound as `DataType::LOB`, so it emits a `HEXTORAW()` literal instead, as `prepareBinary()` already does there.

Correction to my earlier note about CI: `composer-dependency-analyser` does a plain `composer install`, so it stays red in all three until 2.0.2 is tagged and the drivers bump `yiisoft/db` to `^2.0.2`. I will push that bump when you tag.

KalimeroMK:
@Tigrov both review comments are addressed in 6795ad2 and CI is green (82 checks, mergeable).

Could you merge this and tag 2.0.2? The three driver PRs — yiisoft/db-sqlite#432, yiisoft/db-mysql#482, yiisoft/db-oracle#411 — are ready and only waiting on that tag: their `composer-dependency-analyser` failures clear as soon as `UuidValue` exists in a released version.

tests/Db/Expression/Value/Builder/UuidValueBuilderTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $params = [];
+        $result = $builder->build(new UuidValue(self::UUID), $params);
+
+        $this->assertSame(':qp0', $result);
+        $this->assertEquals(
+            [':qp0' => new Param(DbUuidHelper::uuidToBlob(self::UUID), DataType::LOB)],
+            $params,
+        );
+    }
```
Tigrov: This test looks redundant. There are tests in `db-sqlite`, `db-mysql` and `db-oracle`

KalimeroMK: Agreed, removed in e01d85be. The three driver tests cover the override against real databases, so the anonymous subclass here added nothing.

Tigrov (approved):

src/Expression/Value/Builder/UuidValueBuilder.php
```diff
@@ -0,0 +1,50 @@
...
+ * Builder for {@see UuidValue} expressions.
+ *
+ * Binds the UUID as a string parameter in the canonical form, which is what PostgreSQL `uuid` and MSSQL
+ * `uniqueidentifier` columns expect. DBMS that store a UUID as raw bytes, such as MySQL, MariaDB, SQLite and Oracle,
+ * override {@see prepareValue()} to convert the value with {@see DbUuidHelper::uuidToBlob()} and bind it as
+ * {@see DataType::LOB}, so the driver sends it as binary rather than as a character string.
+ *
+ * @implements ExpressionBuilderInterface<UuidValue>
+ */
+class UuidValueBuilder implements ExpressionBuilderInterface
```
vjik: This class is very simple. What about mark it as final and don't use inheritance in DBMS-specific packages?

I know, that many classes in Yii DB are not final and inherited in  DBMS-specific packages. But it increases coupling and fragility.

KalimeroMK: Agreed, done in a757103a.

`UuidValueBuilder` is `final` now and the `protected prepareValue()` hook is gone — `build()` binds the `Param` directly. The three DBMS-specific packages implement `ExpressionBuilderInterface` themselves instead of extending it:

- yiisoft/db-mysql#482
- yiisoft/db-sqlite#432
- yiisoft/db-oracle#411

vjik (approved):

src/Expression/Value/Builder/UuidValueBuilder.php
```diff
@@ -0,0 +1,37 @@
...
+     * @param QueryBuilderInterface $queryBuilder The query builder instance.
+     */
+    public function __construct(
+        private readonly QueryBuilderInterface $queryBuilder,
+    ) {}
+
+    public function build(ExpressionInterface $expression, array &$params = []): string
+    {
+        /** @var UuidValue $expression */
+        return $this->queryBuilder->buildValue(new Param($expression->value, DataType::STRING), $params);
```
Tigrov:
```suggestion
        return $this->queryBuilder->bindParam(new Param($expression->value, DataType::STRING), $params);
```

`bindParam()` is more suitable here.

zoranbogoevskimkd: Applied in 69c141d2 — thanks.

Tigrov:
@KalimeroMK Thanks :+1:

